PR #9, opened 2017-10-09
build: fix out of tree builds

andred:
libtpms.syms is inside the source directory, so when doing out
of tree builds it can't be found - fix the libtool invocation.


stefanberger:
Thanks. Merged.

PR #22, opened 2018-12-13
Improve Debian packaging

strugee:
This squashes some Lintian warnings. Not all of them, but at least a decent amount :thumbsup:


strugee:
Also pushed a commit making the `-dev` package depend on the binary package since that's usually what Debian library packages do.

debian/changelog
```diff
@@ -1,4 +1,4 @@
-libtpms (0.6.0-1-dev1) RELEASED; urgency=medium
+libtpms (0.6.1-1-dev1) RELEASED; urgency=medium
```
stefanberger: Can you also please add a Signed-off-by: to your patches?

debian/control
```diff
@@ -5,7 +5,7 @@ Priority: optional
 Build-Depends: automake, autoconf, libtool, libssl-dev, sed,
  make, coreutils, pkg-config, gawk, sed, bash, dh-exec
 
-Package: libtpms
+Package: libtpms0
```
stefanberger: With this change you also have to rename debian/libtpms libtpms.install to libtpms0.install.

stefanberger: Are you going to make these changes? If not, I will do them myself on Thursday.

debian/control
```diff
@@ -18,5 +18,5 @@ Package: libtpms-dev
 Architecture: all
 Section: libdevel
 Description: libtpms header files and man pages
- The libtpms-dev package provides header files and man pages for the
+ The libtpms0-dev package provides header files and man pages for the
```
stefanberger: This is not correct.

stefanberger:
Thanks for the PR. Though without activity on the requested changes I have to close the PR now and will make those changes myself with reference to you. Thanks.

PR #24, opened 2019-01-03
Upgrade the TPM 2 code from rev 146 to rev 150

stefanberger:
This series of patches upgrades the TPM 2 code from rev 146 to rev 150. Some changes in this series do not allow the downgrading of TPM 2 state anymore since older versions of the TPM 2 wouldn't know how to deal with certain variables.


stefanberger:
PR fixed to compile on all available Linuxs (x86_64, i386, and ppc64) and BSDs.

PR #37, opened 2019-03-27
build-sys: add oss-fuzz support

elmarco:
This script will permit integration with Google OSS-FUZZ
https://github.com/google/oss-fuzz


stefanberger:
LGTM

PR #38, opened 2019-03-29
travis: Use cpp-coveralls with the -b option for proper path creation

stefanberger:
Add the option -b ./src for proper path creation.


stefanberger:
Doesn't work with -b option.

PR #50, opened 2019-04-16
build-sys: don't tweak LD_LIBRARY_PATH

elmarco:
libtool already provides helper scripts around executables.


stefanberger:
I had this there when I was running the test cases from command and didn't want it to pick up the installed libtpms... Ok, let's remove it.

PR #61, opened 2019-05-24
Have the TPM 2 code use OpenSSL's symmetric cipher functions for AES and TDES.

stefanberger:
This set of patches adds an implementation for using OpenSSL's symmetric cipher functions (EVP) for AES and TDES. We leave the original code around, which again can be activate by setting USE_OPENSSL_FUNCTIONS to 'NO'.

### DIFF
--- a/src/tpm2/crypto/openssl/CryptSym.c
+++ b/src/tpm2/crypto/openssl/CryptSym.c
@@ -67,6 +67,7 @@
 /* 10.2.19.2 Includes, Defines, and Typedefs */
 #include "Tpm.h"
 #include "CryptSym.h"
+#include "Helpers_fp.h"  // libtpms changed
 /* 10.2.19.3.1	CryptSymInit() */
 /* This function is called to do _TPM_Init() processing */
 BOOL
@@ -155,6 +156,8 @@ CryptGetSymmetricBlockSize(
 	}
     return 0;
 }
+
+#if !USE_OPENSSL_FUNCTIONS // libtpms added
 /* 10.2.20.5 Symmetric Encryption */
 /* This function performs symmetric encryption based on the mode. */
 /* Error Returns Meaning */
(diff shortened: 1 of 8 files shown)


src/tpm2/crypto/openssl/CryptSym.c
```diff
@@ -458,6 +461,262 @@ CryptSymmetricDecrypt(
...
+        ERROR_RETURN(TPM_RC_FAILURE);
+
+    if (EVP_EncryptInit_ex(ctx, evpfn(), NULL, keyToUse, iv) != 1)
+        ERROR_RETURN(TPM_RC_FAILURE);
+
+    EVP_CIPHER_CTX_set_flags(ctx, EVP_CIPH_NO_PADDING);
+    if (EVP_EncryptUpdate(ctx, pOut, &outlen1, dIn, dSize) != 1)
+        ERROR_RETURN(TPM_RC_FAILURE);
+
+    EVP_CIPHER_CTX_set_flags(ctx, EVP_CIPH_NO_PADDING);
```
simo5: Why are you calling this twice ?

src/tpm2/crypto/openssl/CryptSym.c
```diff
@@ -458,6 +461,262 @@ CryptSymmetricDecrypt(
...
+    if (EVP_EncryptFinal_ex(ctx, pOut + outlen1, &outlen2) != 1)
+        ERROR_RETURN(TPM_RC_FAILURE);
+
+    pAssert(dSize >= outlen1 + outlen2);
```
simo5: It is too late to asser there, you may have already overwritten memory, the check, if you think is needed, should be done beofre calling EVP_EncryptFinal_ex().
Something like this I think:
pAssert(oulen1 == dSize || dSize >= outlen1 + blockSize);

src/tpm2/crypto/openssl/CryptSym.c
```diff
@@ -458,6 +461,262 @@ CryptSymmetricDecrypt(
...
+    UINT16               keyToUseLen = (UINT16)sizeof(keyToUse);
+    TPM_RC               retVal = TPM_RC_SUCCESS;
+
+    // These are used but the compiler can't tell because they are initialized
+    // in case statements and it can't tell if they are always initialized
+    // when needed, so... Comment these out if the compiler can tell or doesn't
+    // care that these are initialized before use.
+    pAssert(dOut != NULL && key != NULL && dIn != NULL);
+    if(dSize == 0)
+	return TPM_RC_SUCCESS;
```
simo5: It is leagl to request decryption of a 0 size buffer ?

src/tpm2/crypto/openssl/CryptSym.c
```diff
@@ -458,6 +461,262 @@ CryptSymmetricDecrypt(
...
+        ERROR_RETURN(TPM_RC_FAILURE);
+
+    if (EVP_DecryptInit_ex(ctx, evpfn(), NULL, keyToUse, iv) != 1)
+        ERROR_RETURN(TPM_RC_FAILURE);
+
+    EVP_CIPHER_CTX_set_flags(ctx, EVP_CIPH_NO_PADDING);
+    if (EVP_DecryptUpdate(ctx, pOut, &outlen1, dIn, dSize) != 1)
+        ERROR_RETURN(TPM_RC_FAILURE);
+
+    EVP_CIPHER_CTX_set_flags(ctx, EVP_CIPH_NO_PADDING);
```
simo5: repeated as in encryption

src/tpm2/crypto/openssl/CryptSym.c
```diff
@@ -458,6 +461,262 @@ CryptSymmetricDecrypt(
...
+    if (EVP_DecryptFinal(ctx, pOut + outlen1, &outlen2) != 1)
+        ERROR_RETURN(TPM_RC_FAILURE);
+
+    pAssert(dSize >= outlen1 + outlen2);
```
simo5: same as for encryption, too late here

src/tpm2/crypto/openssl/CryptSym.c
```diff
@@ -458,6 +461,262 @@ CryptSymmetricDecrypt(
...
+    // If the iv is provided, then it is expected to be block sized. In some cases,
+    // the caller is providing an array of 0's that is equal to [MAX_SYM_BLOCK_SIZE]
+    // with no knowledge of the actual block size. This function will set it.
+    if((ivInOut != NULL) && (mode != ALG_ECB_VALUE))
+	{
+	    ivInOut->t.size = blockSize;
+	    iv = ivInOut->t.buffer;
+	}
+    else
+	iv = defaultIv;
```
simo5: If there is no IV in input, shouldn't this interface generate an IV and return it ?
ivInOut is marked as IN/OUT but I do not see an IV is ever returned

src/tpm2/crypto/openssl/CryptSym.c
stefanberger: Removed. There were some test case failure long ago...

src/tpm2/crypto/openssl/CryptSym.c
stefanberger: This header part with all these tests is a copy of the original function.

src/tpm2/crypto/openssl/CryptSym.c
stefanberger: ... and not necessary in the decryption part.

src/tpm2/crypto/openssl/CryptSym.c
stefanberger: yes...

src/tpm2/crypto/openssl/CryptSym.c
stefanberger: In most cases the IV is a variable on the stack and passed into this function and then not looked at anymore. In some cases the IV comes in via a TPM command directly and is in the input parameters, in which cases it's also not looked at anymore. So I think this IN/OUT is an OUT mostly for efficiency purposes because the buffer of the IV may get changed but seems to not be of interest to the caller afterwards.

src/tpm2/crypto/openssl/CryptSym.c
stefanberger: Right. Thank you.

src/tpm2/crypto/openssl/CryptSym.c
stefanberger: ... and regarding generation of the IV: I am using the same code for the defaultIV as the original code does further up.

stefanberger:
@simo5 Thanks for your comments. I addressed them and some issues raised by Coverity. Unless you have any more concerns, I'd merge this.

simo5:
No other comments, thanks.

PR #65, opened 2019-06-03
Convert to OpenSSL RSA functions for encrypting and signing

stefanberger:
This series of patches converts existing RSA code for encryption and signing to use OpenSSL code for doing these operations.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -186,7 +186,7 @@ AS_IF([test "x$enable_use_openssl_functions" != "xno"], [
 	AC_CHECK_LIB([crypto], [EC_KEY_set_group],, not_found=1)
 	if test "x$not_found" = "x0"; then
 		use_openssl_functions_ecdsa=1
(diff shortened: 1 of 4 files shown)


stefanberger:
Some performance measurements.

Running this loop here (with IBM TSS and talking over sockets to swtpm): `time bash -c "for ((i=0;i<10000;i++)); do ./sign -hk 80000000 -if text -os text.sig; done`

- original TPM 2 code:

```
real	1m17.194s
user	0m23.288s
sys	0m20.126s
```

-  Using OpenSSL and setting CRT parameters for private key:

```
real	2m29.759s
user	0m29.511s
sys	0m24.230s
```

- Using OpenSSL and not setting CRT parameters for private key:

```
real	2m29.676s
user	0m29.264s
sys	0m24.693s

```

simo5:
Looks reasonable to me.

PR #66, opened 2019-06-03
Use OpenSSL function for ECDSA signature creation

stefanberger:
This series of patches converts the ECDSA signing code to use OpenSSL crypto library functions.


stefanberger:
Thanks.

PR #67, opened 2019-06-08
Implement EC key generation using OpenSSL functions if rand == NULL

stefanberger:
Use OpenSSL functions to create EC keys only for the case that
rand == NULL in which case no KDF is being used and where we can
create a truly random key. This doesn't break the upgrade path.


src/tpm2/crypto/openssl/CryptEccMain.c
```diff
@@ -557,6 +562,11 @@ BnEccGetPrivate(
     UINT32                   orderBytes = BITS_TO_BYTES(orderBits);
     BN_VAR(bnExtraBits, MAX_ECC_KEY_BITS + 64);
     BN_VAR(nMinus1, MAX_ECC_KEY_BITS);
+
+#if USE_OPENSSL_FUNCTIONS_EC
```
simo5: I think it would be clearer to have two versions of the function, one that is compiled if USE_OPENSSL_FUNCTIONS_EC and the classic one, however code-wise I see no issues.

stefanberger: Did this now. Will merge. Thanks for the comments.

PR #68, opened 2019-06-10
tpm2: Use OpenSSL to create TDES keys if rand == NULL

stefanberger:
Use OpenSSL function to create TDES keys if rand == NULL,
which indicates that a truely random key needs to be generated
rather than one derived from a KDF.


simo5:
I will take a look at this soon, sorry but TDES has lower priority because NIST is deprecating its use completely.

stefanberger:
@simo5 If you have a chance to review, please do, otherwise I would merge it later today ...

PR #70, opened 2019-06-18
WIP: Update the code to revision 155

stefanberger:
This series of patches updates the code to revision 155. Support for the TPM 2 creating X509 certificates is being added here.

One complication is the new CryptAdjustPrimeCandidate() function that creates slightly different prime numbers than the original code and therefore causes derived keys to be different now, which can have the side effect that keys cannot be used anymore for decryption. For this we have to start tracking the 'age' of the seeds that are at the core of the key derivation functions and only use the new CryptAdjustPrimeCandidate() function if a new seed is being used. A new seed is being used when either a completely new TPM is created or if one of its seeds is renewed using one of the TPM 2 commands TPM2_ChangeEPS, TPM2_ChangePPS, or TPM2_Clear.

The new CryptAdjustPrimeCandidate() function solves an issue where the old code created slighly weaker keys on 64 bit machines since it set some bytes to 0. The new code, however, (still) created slightly different prime numbers for 64 bit and 32 bit machines, which I adjusted so that the 32 bit algorithm is used and now the same prime numbers are created on 64 bit and 32 bit machines when the same seed is used.


stefanberger:
Closing this one in favor of #96

PR #71, opened 2019-06-19
tpm2: Use OpenSSL for decryption and signature verification

stefanberger:
Use OpenSSL funtions now also for decryption and signature verification.


stefanberger:
@simo5 Hi Simo! I may need your help with constant time. Are you familiar with it?

simo5:
@stefanberger yes, I can help, tag me as needed

stefanberger:
I have #163 now. There's a text file where I tried to assess the OpenSSL function that react to BN_FLG_CONSTTIME and then below that is a part with all the OpenSSL BIGNUM functions called from the TPM 2 and where I think I should make changes to. I think I got most of them right for RSA.Please have a look and let's discuss over there then.

PR #96, opened 2019-12-13
Update TPM 2 code to revision 159

stefanberger:
This series of patches upgrades the TPM 2.0 code to revision 159.

The TCG changes have been broken apart into smaller, easier to understand, chunks. 

Revision 155 made a fix to the RSA prime number generation algorithm. The libtpms code needs to handle this in a backwards compatible way by using the old key generation algorithm that derives an RSA key from the seed. Once the seed is re-newed, the new algorithm may be used. So, there's an identifier now attached to each seed that allows us to determine whether the seed is old (and old key generation algorithm must be used) or new.

Rev. 155 also added X509 certificate creation to the code. This code has been enabled.

Rev. 159 added Authenticated Timers to the code. Those are entirely disabled.


src/tpm2/StartupCommands.c
```diff
@@ -320,13 +320,10 @@ TPM2_Shutdown(
 	    else if(g_StartupLocality3)
 		gp.orderlyState = TPM_SU_STATE | STARTUP_LOCALITY_3;
 	}
-    else if(in->shutdownType == TPM_SU_CLEAR)
+    else if(in->shutdownType != TPM_SU_CLEAR)
 	{
-	    // Save STATE_RESET data
-	    NvWrite(NV_STATE_RESET_DATA, sizeof(STATE_RESET_DATA), &gr);
```
stefanberger: Check that NV_STATE_RESET_DATA need not be saved...

stefanberger: .... this seems correct.

PR #115, opened 2020-01-31
tpm2/build-sys: Add PC Client compliant tpm2

joholl:
Most TPM2s in consumer PCs are compliant to the TCG PC Client Platform TPM Profile (PTP) Specification. To be able to test applications against a TPM with minimal pc client feature set, I added a new configure option `--with-pcclient`. If this option is given, an *additional* library `libtpms-pcclient` is built. This PR is very much a proof of concept and I did not touch the tests, yet

To use this library with swtpm, you currently have to use `LD_PRELOAD`. In the end, I would like an swtpm argument (e.g. `--pcclient` or `--minimal`) to `dlopen()` this library. Alternatively, one could configure the libtpms library at compile time. I guess, this would be easier to implement but less comfortable to use.

What do you think?


stefanberger:
Thanks for the PR. In terms of ordering, your PR would have to come in after PR #96, which also does some major refactoring, though I would like to understand your changes first. The fact that there is another library unfortunately does not make things easier. I suppose most differences are in the values for #defines, but what are they and what effect do they have on client apps / TSS2?

stefanberger:
Ok, I see the changes. One concern is testing of this configuration. Another is the potential mixup with states of existing libtpms with libtpms-pcclient, which are of course incompatible. I am not sure how we will maintain two different version and possibly more.

joholl:
Sorry for the late response, I wanted to implement testing first :)

> In terms of ordering, your PR would have to come in after PR #96, which also does some major refactoring

That'd be ok for me. Just let me know when to rebase.

> One concern is testing of this configuration. Another is the potential mixup with states of existing libtpms with libtpms-pcclient, which are of course incompatible. I am not sure how we will maintain two different version and possibly more

I see your concerns. To be honest, I first also had problems with having the wrong states loaded. I added  ba4946a which appends pcclient states with `-pcclient`. That should fix it.

As for testing: I added some Automake magic to test the pcclient configuration. Basically, if `--with-pcclient` was passed, every test binary is built twice, the original linking against `libtpms.so` and the pcclient version linking against `libtpms-pcclient.so`. I also added a travis build for reference.

The shell tests were a bit tricky: each test script decides upon its name which test binary to run. I.e. `tpm2_createprimary.sh` will call `tpm2_createprimary` whereas `tpm2_createprimary.pcclient` will call `tpm2_createprimary_pcclient`. Then again, `tpm2_createprimary_pcclient` is only a symlink to `tpm2_createprimary.sh` and built when `make check` is called, so no additional code.

Note that the code base of both `libtpm` and `libtpms-pcclient` is the same. The only difference is the `Implementation.h` file. Plus, the test code is the same as well, that should keep the maintanance effort minimal.

stefanberger:
So the concern with the testing is that the TPM 2 state has to tolerate a code upgrade as well as possibly configuration upgrade. There are several tests with state in the swtpm project that resume the TPM with state that was created a long time ago and it still works. Ideally we would have some tests for this type of state there as well.

Though generally I am wondering whether it wouldn't be sufficient to only have the current vTPM configuration that results in a superset of (crypto) functionality compared to the PC client  configuration.

joholl:
I agree, testing the fully featured configuration should ensure that none of the functionality breaks. The pc client conf is just a subset of these features. If at all, we might want to add a test for preventing the complete removal of a feature in the pc client conf which would brak loading old states.

joholl:
I'll keep you posted while I implement a PoC for the swtpm. This may take a while :)

stefanberger:
> I'll keep you posted while I implement a PoC for the swtpm. This may take a while :)

Thanks. I am also concentrating on something else currently.

For swtpm, I think there we should have a command line parameter like `--profile name=pcclient`. assuming the default profile if left empty. Likely we should extend the state header to include the name of the profile with an empty profile for the default so we can refuse starting the TPM if the user chooses the wrong profile. We'll need to bump up the header to the next version and then also have the header be part of the HMAC. The test cases may be a bit more complicated to extend.

src/tpm2/ImplementationPCClient.h
```diff
@@ -0,0 +1,1222 @@
...
+#ifndef  BIG_ENDIAN_TPM
+#define  BIG_ENDIAN_TPM       NO
+#endif
+#define  LITTLE_ENDIAN_TPM          !BIG_ENDIAN_TPM
+#define  MOST_SIGNIFICANT_BIT_0     NO
+#define  LEAST_SIGNIFICANT_BIT_0    !MOST_SIGNIFICANT_BIT_0
+#define  AUTO_ALIGN                 NO
+
+/* From TCG PC Client Platform 01.03v22: Table 2 PC Client TPM Algorithms */
+#define  RSA_KEY_SIZES_BITS         {2048}
```
stefanberger: I heared about 3072 bits coming up. While this is not a problem of the pcclient config alone, I am wondering how to handle the bump to the next sizes.

joholl: The PC Client config is supposed to be **minimal**, that means any optional features are disabled. Thus, if 3072 is not mandatory, it will be disabled in `ImplementationPCClient.h.`

joholl:
> For swtpm, I think there we should have a command line parameter like --profile name=pcclient. assuming the default profile if left empty. Likely we should extend the state header to include the name of the profile with an empty profile for the default so we can refuse starting the TPM if the user chooses the wrong profile. We'll need to bump up the header to the next version and then also have the header be part of the HMAC. The test cases may be a bit more complicated to extend.

To be honest, I'm not sure if this is really necessary. I'd suggest simply postfixing the state files' names with e.g. `-pcclient` (as implemented in my latest PR). In this case, a profile conflict can only occur if **a)** someone specificly renames the state file and switches profiles or **b)** deliberately loads the state file of another profile.

Renaming the state files with regard to the profile seems lika a sufficient countermeasure for avoiding accidental profile conflicts to me.

https://github.com/stefanberger/libtpms/blob/ba4946a2d447232240fbe6505bb4c6fb5a717738/include/libtpms/tpm_nvfilename.h#L45-L59

src/tpm2/ImplementationPCClient.h
stefanberger: It will be bumped up to 3072 in the next version...

joholl: Oh, you are right, I missed that. This PR is based on v1.03 anyway. Since the pcclient simulator will probably mainly be used for compatibility testing, less features is generally better (for now).

stefanberger: we will have to base it on the latest accepted version...

joholl: I think it is just as valid to support older versions. However, since I do not really have a strong opinion about this, I can update the values to v1.04 if you wish. I'll keep you posted.

stefanberger: The new version is not far away and it doesn't make sense that the pcclient profile would be immediately obsoleted.

joholl:
Sorry that it took so long. Here the new changes:

- bumped PC Client version up to 01.04rev37
- reworked the configure magic, now the options are:
   - `--with-extra-profiles=pcclient`: Here you can specify a list of profiles you want to build (besides the default one)
   - `--with-extra-profile-tests=pcclient`: Here you can specify **one** profile which will be tested additionally to the default one (mostly for CI)

Have also a look at stefanberger/swtpm#215 where I added an option `--profile name=<name>` to load a specific profile.

stefanberger:
I am not sure when I will be able to take this. There's quite some complexity associated with the state of a profile and particulaly upgrading it and of course there need to be **all the tests on the swtpm level** as well.

joholl:
Hm, I thought since the `pcclient` profile is essentially a subset of the `default` profile, it should not be too much of a deal, even when updating. In the end, the only difference are some defines in `Implementation.h`.

I worked on the swtpm tests a little bit and most of them succeed after minor adjustments. Currently, I have 8 failing tests. 
* `test_tpm2_swtpm_setup_create_cert`: calls `swtpm_setup` which issues `tpm2_changeeps` (not supported on `pcclient`)

The rest of the failing tests seems to load state files from `tests/data`
* `test_tpm2_ctrlchannel2`
* `test_tpm2_encrypted_state`
* `test_tpm2_migration_key`
* `test_tpm2_resume_volatile`
* `test_tpm2_save_load_encrypted_state`
* `test_tpm2_volatilestate`
* `test_tpm2_save_load_state_da_timeout`

Since the `pcclient` is just a feature subset, I'm not sure if we need to run all these test against the the `pcclient` profile, as well. What do you think?

About creating states: I did not find a way to get volatile states for tpm2. Also, I have a hard time figuring out how to create the states I need. Did you create the states manually? Is there any doc besides the test code? 

It would be really great to have a minimally featured PC Client compliant TPM simulator to test software against. Currently, there is no proper way of preventing _command not supported_ errors.

P.S.: [the TSS is using swtpm](https://github.com/tpm2-software/tpm2-tss/pull/1542)  :)

stefanberger:
The issue is mostly the state but also making sure upgrades of the TPM code still work for a particular profie. The state libtpms writes out on day one has to still be readable and usable after all upgrades of libtpms even when new features were added to the profile or simply new code was added. That's why there are all these tests also with state created by older versions of libtpms. This is the biggest challenge of this project and any new profile adds complexity to it.

I have to look at your patches again. But one particular thing I seem to remember is the change to the code (#ifdef'd out) in NVMarhsal.c related to some data structure that pcclient doesn't have. It would diverge the code and become really bad if one day the pcclient profile was to enable it. I think we need to bump up the version of the structure where this is in and write a byte into the byte stream (which is available only starting with the 'bumped up version' once we get to read the bytestream) indicating whether the sub-data structure is available or not. In pcclient case it's not available and no further data are read and in the current profile case the data are then read via a function call (if I remember correctly).

joholl:
I see. So we'd need to create "day one states" and tests for loading them. I took a look at the states for the default profile, but I found no indication of how to create these states. Can you point me in the right direction? Also, I wonder how to create volatile states.

To be honest, I'm not familiar which exact sub-data structure we're talking about and how to change the state structure in a sensible way. I'd need your assistance in that one, too.

Another solution might be not supporting loading old pcclient states, at all. Using non-default profiles could be an unsupported feature as in "use at your own risk". It will most likely used for testing only, anyway (which does not involve saving states). Adding a disclaimer to the `configure` help page or a warning in the config log should be sufficient. 

This does not mean that we should omit testing. Most tests succeed already, I'm only talking about the loading-old-states issue.

What do you think?

stefanberger:
> Another solution might be not supporting loading old pcclient states, at all. Using non-default profiles could be an unsupported feature as in "use at your own risk". It will most likely used for testing only, anyway (which does not involve saving states). Adding a disclaimer to the `configure` help page or a warning in the config log should be sufficient.

I am hesitant to do this. If someone uses the pcclient profile he and every data center where it's used would expect the state to be loadable again after an upgrade of libtpms. So somehow we would have to prevent it from being packaged then since it's basically unsupported. If someone wanted to just patch libtpms and rebuild libtpms locally to have a pcclient compliant by version, then that's something everyone can do on their own and we don't need official support for this.

If we wanted to support another profile now (and I am not sure we should since the TCG profile updates are enabling more and more features) we should probably have the following design. 

libtpms would have to do the `dlopen` in TPMLIB_MainInit() rather than having swtpm doing this. There should be another API call TPMLIB_ChooseTPM2Profile(const char *) so select TPM 2 and its profile, NULL or 'default' being the default profile. The interface functions that the 'profile' libraries libtpms-tpm12.so, libtpms-tpm2-default.so, and libtpms-tpm2-pcclient.so would offer are described by this existing interface here:

https://github.com/stefanberger/libtpms/blob/master/src/tpm_library_intern.h#L61
```
struct tpm_interface {
    TPM_RESULT (*MainInit)(void);
    void (*Terminate)(void);
    uint32_t (*SetBufferSize)(uint32_t wanted_size, uint32_t *min_size,
                              uint32_t *max_size);
    TPM_RESULT (*Process)(unsigned char **respbuffer, uint32_t *resp_size,
                          uint32_t *respbufsize,
		          unsigned char *command, uint32_t command_size);
    TPM_RESULT (*VolatileAllStore)(unsigned char **buffer, uint32_t *buflen);
    TPM_RESULT (*CancelCommand)(void);
    TPM_RESULT (*GetTPMProperty)(enum TPMLIB_TPMProperty prop,
                                 int *result);
    char *(*GetInfo)(enum TPMLIB_InfoFlags flags);
    TPM_RESULT (*TpmEstablishedGet)(TPM_BOOL *tpmEstablished);
    TPM_RESULT (*TpmEstablishedReset)(void);
    TPM_RESULT (*HashStart)(void);
    TPM_RESULT (*HashData)(const unsigned char *data,
                           uint32_t data_length);
    TPM_RESULT (*HashEnd)(void);
    TPM_RESULT (*ValidateState)(enum TPMLIB_StateType st,
                                unsigned int flags);
    TPM_RESULT (*SetState)(enum TPMLIB_StateType st,
                           const unsigned char *buffer, uint32_t buflen);
    TPM_RESULT (*GetState)(enum TPMLIB_StateType st,
                           unsigned char **buffer, uint32_t *buflen);
};
```

So this would mean refactoring libtpms and building 3 shared libraries at the moment with the currently built one dynamically loading the other two in TPMLIB_ChooseTPM2Profile() or TPMLIB_MainInit(). Probably the latter.

Swtpm would just call the new API call and have an option for getting the profile from the user. All test cases in swtpm project should have the profile name being passed as environment variable so that we can run the entire test suite with either one of the profiles. A similar case is the passing of  SWTPM_TEST_SECCOMP_OPT except that in this case we may just want to pass the plain profile name and then use something like this here to formulate the option on the command line:

``` SWTPM_TEST_TPM2_PROFILE=x; echo ${SWTPM_TEST_TPM2_PROFILE:+--profile $SWTPM_TEST_TPM2_PROFILE} ```

Some test cases will have to be adapted to provide state files for both profiles and work with different 'constants' depending on which profile is chosen. I could then handle the creation of the state files or would have to direct you how to do this. Those state files would obviously have to be created by using libtpms-tpm2-pcclient.so.

As always, time is an issue for me.

joholl:
I looked into this, but I could not properly separate the code of main, tpm2 and tpm12 libraries within a reasonable amount of time. There is quite some code e.g. in tpm12 used by tpm2 etc. To properly separate these libraries to a point where there is only a dependency from main to tpm12 and tpm2 is too much work for me.

I hoped to have different entire `libtpms.so` libraries and have swtpm link against them dynamically. If your suggested approach is the only viable solution to you, feel free to close this issue (and the related issue at your swtpm repository).

Either way, thank you for your time and effort up to now!

joholl:
This PR has gone stale.

PR #116, opened 2020-04-03
fix debian/{control,changelog}

realtime-neil:


debian/control
```diff
@@ -2,8 +2,15 @@ Source: libtpms
...
- gawk, dh-exec, debhelper (>= 9), g++
+Standards-Version: 3.9.7
+Build-Depends:
+ autoconf,
+ automake,
+ debhelper (>= 9),
+ dh-exec,
+ gawk,
+ libssl-dev,
+ libtool
```
stefanberger, line 13: g++ is needed for the fuzzing test.

realtime-neil: > g++ is needed for the fuzzing test.

You're correct, and I should have mentioned that initially. I would argue that the fuzzing test and the Debian package build are two different things that require similar --- but not identical --- dependencies. That the `.travis-ci.yml` key `addons.apt.packages` nearly duplicates the `debian/control` `Build-Depends` seems a tacit admission of this fact.

realtime-neil: @stefanberger , also your statement begs the question: How are you currently using `debian/control` to prepare an environment for any fuzzing test? If you're using `mk-build-deps` to accomplish this, then I can't find it.

realtime-neil: @stefanberger , in the interest of better understanding your position, let me start by constructing a straw man. Let's assume for the moment that you _were_ using the `debian/control` file to install packages needed by the fuzz test. In this hypothetical, you're using `mk-build-deps` via tooling absent from source control, or some other, subtler mechanism in source control that I can't seem to find. The former would be a highly unconventional use of that tool, while the latter would be a project-specific `debian/control` parser that has eluded my detection.

Even assuming such mechanisms existed and worked as described, we're still talking about conflation of dependency sets; i.e., the implicit assumption that building the Debian archive and fuzz-testing the binaries should require the same things. Would you not agree that this assumption is dubious at best? Would you not agree that fuzz-testing requires things that packaging does not --- and vice-versa? If we keep pushing this straw man, the result is a Debian package containing binaries instrumented for fuzz-testing. I can't quote chapter and verse, but I suspect this would be a departure from the Debian policy.

stefanberger: No need to construct a straw man. I'll merge it later.

realtime-neil:
@stefanberger you're fuzz-testing as part of the Debian archive build? Or you're building and packaging things that will drive fuzz-testing?

stefanberger:
It was fuzzing while testing what it built. It fuzzed libtpms.

realtime-neil:
@stefanberger would you consider a change that moves the `make` target responsible for fuzz-testing out of the `check` target? What do you have for test units that _don't_ do fuzzing?

stefanberger:
Everything is running under `make check` right now. Fuzzing should probably removed from it - at least once we have more test vectors in the repo and the fuzzing starts taking a long time.

realtime-neil:
Huh. Okay. Welp, color me informed. My experience with fuzz-testing is minuscule, so I'll wait and watch.

stefanberger:
If you just add g++ to the list of dependencies, I'd merge.

realtime-neil:
> If you just add g++ to the list of dependencies, I'd merge.

And fail the `lintian` checks I got passing? Nah. This unit testing situation is more pressing -- I'll take a look at that later in the week.

stefanberger:
Ah, `g++` is part of built-essential. Merging. Thanks!

realtime-neil:
Woah! Thank _you_!

stefanberger:
Maybe that at some point back I didn't have build-essential installed (??) or something had happend to g++ and I ended up adding it explicitly... Anyway, thanks.

PR #117, opened 2020-04-06
make debhelper play nice with autotools

realtime-neil:
We need to inform debhelper that it will be driving the action using autotools.

edit `debian/control` field `Build-Depends`: add `autotools-dev` and
`dh-autoreconf`

edit `debian/rules`:

* change the default target `%` to invoke `dh` with `autotools-dev` and
  `autoreconf`

* remove from `override_dh_auto_configure` the invocation to `./autogen.sh`
  obsoleted by `dh-autoreconf`

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -6,7 +6,9 @@ Standards-Version: 3.9.7
 Build-Depends:
  autoconf,
  automake,
+ autotools-dev,
  debhelper (>= 9),
+ dh-autoreconf,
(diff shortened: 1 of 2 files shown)


stefanberger:
Thank you!

realtime-neil:
> Thank you!

My pleasure, good sir. Thank you for writing this in the first place!

Issue #121, opened 2020-04-17
Allow RSA keys to be 3072 bits long


stefanberger:
Modify the TPM 2 code to allow for RSA keys with 3072 bits.

stefanberger:
Modifying the code may involve:
- increase TPM_CONTEXT_SIZE to 2680 in TpmTypes.h
- enabling RSA_3072 in TpmProfile.h

libtpms should also advertise the max size of its RSA keys so that swtpm and possibly swtpm_setup can also advertise it.

stefanberger:
Since the size of OBJECT increases and OBJECTs can be located in NVRAM (TPM_HT_PERSISTENT), we need to expand the NVRAM by ~44 kb, so that an NVRAM filled up with 2048 bit RSA keys will still fit after an upgrade of libtpms and the expansion of OBJECT. The user will gain free NVRAM...

PR #123, opened 2020-04-17
Add support for RSA 3072

stefanberger:
Add support for RSA 3072 keys. This PR addresses issue #121.


tests/nvram_offsets.c
```diff
@@ -59,7 +59,13 @@ int main(void)
      * size of the OBJECT is the same on all architectures so that a full
      * NVRAM fits on all architectures
      */
-#define OBJECT_EXP_SIZE 1896
+#if RSA_4096
+# error Unsuported RSA key size
+#elif RSA_3072
+# define OBJECT_EXP_SIZE 2600
```
stefanberger, line 65: Need to investigate why this increases by so much between RSA_2048 and RSA_3072.

stefanberger: Parts of the OBJECT structure increase by 128, 320, and 256 bytes respectively, thus a total of 704 bytes increase (37%).

PR #174, opened 2021-01-08
tpm2: Support compile-time enablement of HLK compliance

stefanberger:
Some HLK 2004 tests will fail if the NIST P521 curve is supported
by the TPM 2. This patch allows to compile-time disable this.


stefanberger:
Please also see #175 .

stefanberger:
Deferring this for now. Closing it.

PR #175, opened 2021-01-09
Extend API to support dynamic selection of HLK compliance

stefanberger:
This PR extends the API to support a dynamic selection of HLK compliance (certain tests otherwise run into issues with NIST P521 curve if it is supported by a TPM 2). New API calls allow to set, get, and clear flags. One new flag is defined that handles the HLK compliance.


stefanberger:
I am wondering whether HLK compliance means deactivating NIST P521 for HLK 2004 but may for some later HLK version mean something completely different. HLK 2004 may as well just have a bug related to NIST P521.

stefanberger:
Deferring this for now and closing it.

PR #176, opened 2021-02-08
configure: support --with-pkgconfigdir

williamcroberts:
Support setting different install paths for package config files
using the --with-pkgconfigdir option.

Signed-off-by: William Roberts <william.c.roberts@intel.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: c
-dist: trusty
+dist: xenial
 env:
   global:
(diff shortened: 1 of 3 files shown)


williamcroberts:
> Hrm, this didn't seem tp update the install path

Oh it's hardcoded in the Makefile... rm that!

Also no macros below AC_OUTPUT, so move above.

stefanberger:
@williamcroberts Thanks for this PR. Unfortunately there's now a build error on Ubuntu Trusty: https://travis-ci.com/github/stefanberger/libtpms/jobs/481378107

stefanberger:
We could use Xenial as a minimum instead. Maybe that would make it work then.

williamcroberts:
> We could use Xenial as a minimum instead. Maybe that would make it work then.

The macro PKG_INSTALLDIR is not provided until 

> We could use Xenial as a minimum instead. Maybe that would make it work then.

I wonder if this is some bug or something, because the macro and support seem to be since pkgconf-0.8.10, and Trusty seems to be on version 0.94. However, digging up the package it seems the macro is missing.

I guess the question is, do we care about something as old as Trusty and can we bump to xenial? I verified it's in xenial properly, but we also probably want to explicitly pull in pkgconf or it's not there.

williamcroberts:
Alternatively we can package the pkg.m4 macro file into the m4 directory.

stefanberger:
I would just require Xenial as a minimum. I think swtpm requires that also or even later.

Issue #195, opened 2021-02-26
TPM2_ContextSave on public keys loaded using TPM2_LoadExternal leads to a TPM failure


diabonas:
**Describe the bug**
libtpms 0.8.0 since commit 1e1648fec31bd8d80b2ece4eaefe851e2cf4009d uses `ANY_OBJECT_Marshal` to save key contexts. However, public keys loaded using `TPM2_LoadExternal` have a [sensitive portion of `NULL`](https://github.com/stefanberger/libtpms/blob/921d6dc92e153722e3d186dbb92fc2be589e1c38/src/tpm2/ObjectCommands.c#L187), leading to the following error when trying to run `TPM2_ContextSave` on them:
```
libtpms/tpm2: Entering failure mode; code: 4, location: TPMU_SENSITIVE_COMPOSITE_Marshal line 2126
libtpms/tpm2: TPM2_Process: Entered failure mode through command:
80 01 00 00 00 0e 00 00 01 62 80 00 00 00 
```

**To Reproduce**
1. Install libtpms 0.8.0, swtpm 0.5.2, OpenSSL 1.1.1j and [tpm2-tools](https://github.com/tpm2-software/tpm2-tools/) 5.0.
2. Run the following commands:
    ```
    openssl genrsa -out priv.key 2048
    openssl rsa -in priv.key -pubout -out pub.key
    swtpm socket --tpm2 --server port=2321 --ctrl type=tcp,port=2322 --flags not-need-init,startup-clear --tpmstate dir=/tmp &
    tpm2_loadexternal --tcti=swtpm --key-algorithm=rsa --public=pub.key --key-context=key.ctx
    ```
3. Observe the last command failing with the error message
    ```
    libtpms/tpm2: Entering failure mode; code: 4, location: TPMU_SENSITIVE_COMPOSITE_Marshal line 2126
    libtpms/tpm2: TPM2_Process: Entered failure mode through command:
    80 01 00 00 00 0e 00 00 01 62 80 00 00 00
    WARNING:esys:src/tss2-esys/api/Esys_ContextSave.c:247:Esys_ContextSave_Finish() Received TPM Error
    ERROR:esys:src/tss2-esys/api/Esys_ContextSave.c:92:Esys_ContextSave() Esys Finish ErrorCode (0x00000101)
    ERROR: Esys_ContextSave(0x101) - tpm:error(2.0): commands not being accepted because of a TPM failure
    ERROR: Unable to run tpm2_loadexternal
    ```
[...]

PR #196, opened 2021-02-26
tpm2: Fix public key context save due to ANY_OBJECT_Marshal usage

stefanberger:
This patch addresses the bug reported in issue #195 where the saving of
an externally loaded public key's context doesn't work due to the usage of
ANY_CONTEXT_SAVE for saving key contexts. This patch fixes the issue by
creating local versions of TPM_SENSITIVE_Marshal/_Unmarshal that deals
with the case where sensitiveType is not a type of private key but a
public key instead that basically doesn't have much information in
TPM_SENSITIVE but is all zeros instead.

### DIFF
--- a/src/tpm2/NVMarshal.c
+++ b/src/tpm2/NVMarshal.c
@@ -2316,6 +2316,67 @@ skip_future_versions:
     return rc;
 }
 
+/* Local version of TPMT_SENSITIVE_Marshal handling public keys that don't have much in TPM_SENSITIVE */
+static UINT16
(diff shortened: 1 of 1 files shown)


diabonas (approved):
Thank you very much for the quick fix! I confirm that this pull request fixes the minimal reproducer given in #195, as well as passing all tpm2-software test suites (including the previously failing tpm2-tools tests).

stefanberger:
So great! It actually uncovered a much bigger issue in libtpms...

PR #205, opened 2021-03-02
Fix many misspellings

niooss-ledger:
When testing downgrading from libtpms 0.8 to 0.7 (which is not possible), the error message which is reported is:

    libtpms/tpm2: Unexpect value for MAX_RSA_KEY_BITS; its value 3072 is
    not = 2048; (version: 2).

`codespell` (https://github.com/codespell-project/codespell) reports a misspelling for "Unexpect", which should be "Unexpected". As the project contains many more misspellings in comments, error messages and documentation, fix all misspellings reported by `codespell`.


stefanberger:
Thanks!

PR #216, opened 2021-05-18
build-sys: leave CFLAGS/LDFLAGS for user to be defined

elmarco:
This allows user to set specific flags during compilation, without
overriding configure-time cflags necessary for compilation.

See also:
https://www.gnu.org/software/automake/manual/html_node/User-Variables.html
https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html


stefanberger:
Something is wrong with the coverage now.

PR #223, opened 2021-06-21
Reset buffer size indicators that are too large and check for maximum size on marshalling

stefanberger:
Reset buffer size indicators that are found to be too large so that we don't have bad buffer sizes in memory that can cause issues when marshaling the data upon saving volatile data for example.
Check for the maximum size of a buffer when marshaling data so we do not overstep the TPM2B's buffer size.

### DIFF
--- a/src/tpm2/Marshal.c
+++ b/src/tpm2/Marshal.c
@@ -59,6 +59,7 @@
 /*										*/
 /********************************************************************************/
 
+#include <assert.h> // libtpms added
 #include <string.h>
 
 #include "Tpm.h"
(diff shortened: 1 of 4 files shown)


elmarco:
Isn't there anything better than `assert()` that could be done? If it can be triggered from the guest, it may still be exploitable to break the VM in some ways.

stefanberger:
@elmarco What do you suggest?

elmarco:
If the assert can't be triggered at runtime (which you told me it shouldn't), it is fine.

stefanberger:
Yes, that would be a violation of the buffer size...

PR #234, opened 2021-07-22
tests: Extend the fuzzer with state suspend/resume at every step

stefanberger:
This PR extends the fuzzer with state suspend/resume at every step.

### DIFF
--- a/tests/fuzz.cc
+++ b/tests/fuzz.cc
@@ -9,6 +9,13 @@
 #include <libtpms/tpm_error.h>
 #include <libtpms/tpm_memory.h>
 
+
+static void die(const char *msg)
+{
+    fprintf(stderr, "%s", msg);
+    assert(false);
+}
+
 extern "C" int LLVMFuzzerTestOneInput(const uint8_t* data, size_t size)
 {
     unsigned char *rbuffer = NULL;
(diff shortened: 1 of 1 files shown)


stefanberger:
@elmarco Any comment?

elmarco:
Any idea of the slowdown? if it's marginal, I would say go for it. If not, perhaps it should be optional. Tbh, I have no idea how well the fuzzing coverage is doing today.

stefanberger:
I think the slow-down is marginal also since we're not touching the filesystem anymore.
Have a look at this here now: https://github.com/stefanberger/libtpms-fuzz-corpus/blob/e795d10b80c127ab9c14d53593de7abcc1006035/Readme.md

elmarco:
> I think the slow-down is marginal also since we're not touching the filesystem anymore.
ok

> Have a look at this here now: https://github.com/stefanberger/libtpms-fuzz-corpus/blob/e795d10b80c127ab9c14d53593de7abcc1006035/Readme.md

This is not actually fuzzing, right? It's just running the fuzz corpus.

You should credit the initial corpus like I did in commit ef356d0ddb10e91dea23dac53167ad57731b20dd imho.

Also, how did you populate the corpus? There are various metrics to observe when growing a corpus, the bigger is not necessarily the better for what I know.

stefanberger:
> This is not actually fuzzing, right? It's just running the fuzz corpus.

Yes, it's only for feeding the corpus into swtpm.

> 
> You should credit the initial corpus like I did in commit [ef356d0](https://github.com/stefanberger/libtpms/commit/ef356d0ddb10e91dea23dac53167ad57731b20dd) imho.
> 
Sure, I can delete the project and start over. Should be ok there.

> Also, how did you populate the corpus? There are various metrics to observe when growing a corpus, the bigger is not necessarily the better for what I know.

I am running the libtpms/test/fuzz and have it create more data. The corpus keeps on growing the longer the fuzzer works its way through the TPM 2 code. It's hard to say whether some of the commands actually add something 'new' to it but it's fairly quick to feed all of these packets into a swtpm instance, so I don't worry too much about that.

elmarco:
> 
> I am running the libtpms/test/fuzz and have it create more data. The corpus keeps on growing the longer the fuzzer works its way through the TPM 2 code. It's hard to say whether some of the commands actually add something 'new' to it but it's fairly quick to feed all of these packets into a swtpm instance, so I don't worry too much about that.

If you read commit ef356d0 message, it explains how that initial corpus was built, and minimized. Imho, it's needless and probably wasteful to add whatever is generated when running the fuzzer without considerations on coverage for example.

stefanberger:
I build the corpus using
```
./fuzz -jobs=4 -workers=4 ./corpus-execute-command/
```
It does not continuously write out new test data but only a few per hour. My guess is that it somehow finds new corners in the code it hasn't reached before and then writes the new test data out.

Regarding performance: The test run  (time ./fuzz.sh) went down from ~31s to 25s. (now you can approx. derive how old my test machine is...)

stefanberger:
I fixed that other project now.

PR #241, opened 2021-07-27
Enable Camellia symmetric encryption

stefanberger:
Enable Camellia symmetric encryption


stefanberger:
Todo:
- Enable 192bit
- Check that Camellia is tested as part of self-test
- Check that Camellia keys are marshaled/unmarshalled properly
- Check for any more compile-time constants to add to the array

Issue #259, opened 2021-09-09
TPM failure when running TPM2_PCR_Event through tpm2-abrmd


diabonas:
**Describe the bug**
Since libtpms version 0.8.5, attempting to run `tpm2 pcrevent` (from [tpm2-tools](https://github.com/tpm2-software/tpm2-tools)) through [tpm2-abrmd](https://github.com/tpm2-software/tpm2-abrmd) generates the following TPM failure:
```
libtpms/tpm2: Entering failure mode; code: 3, location: CryptDigestUpdate line 429
libtpms/tpm2: TPM2_Process: Entered failure mode through command:
80 02 00 00 00 22 00 00 01 5c 80 00 00 00 00 00
00 09 40 00 00 09 00 00 00 00 00 00 05 74 65 73
74 0a
```
I guess this is related to commit https://github.com/stefanberger/libtpms/commit/33a03986e0a09dde439985e0312d1c8fb3743aab, which seems to corrupt the internal TPM state in this use case.

**To Reproduce**
1. Start swtpm and tpm2-abrmd:
    ```
    swtpm socket --tpm2 --server port=2321 --ctrl type=tcp,port=2322 --flags not-need-init --tpmstate dir=/tmp &
    tpm2-abrmd --session --dbus-name=com.intel.tss2.Tabrmd2321 --tcti=swtpm:port=2321 &
    ```
2. Try to run `tpm2 pcrevent` using the simulator:
    ```
    $ echo test | tpm2 pcrevent --tcti="tabrmd:bus_type=session,bus_name=com.intel.tss2.Tabrmd2321"
    WARNING:esys:src/tss2-esys/api/Esys_SequenceUpdate.c:287:Esys_SequenceUpdate_Finish() Received TPM Error
    ERROR:esys:src/tss2-esys/api/Esys_SequenceUpdate.c:97:Esys_SequenceUpdate() Esys Finish ErrorCode (0x00000101)
    ERROR: Esys_SequenceUpdate(0x101) - tpm:error(2.0): commands not being accepted because of a TPM failure
    WARNING:esys:src/tss2-esys/api/Esys_FlushContext.c:234:Esys_FlushContext_Finish() Received TPM Error
    ERROR:esys:src/tss2-esys/api/Esys_FlushContext.c:89:Esys_FlushContext() Esys Finish ErrorCode (0x00000101)
    ERROR: Esys_FlushContext(0x101) - tpm:error(2.0): commands not being accepted because of a TPM failure
[...]

PR #260, opened 2021-09-09
tpm2: Marshal event sequence objects' hash state

stefanberger:
Event sequence objects were never properly marshalled and when their state
was saved and later restored their state may have been corrupted. Fix this
now by also marshalling the state of event sequence objects.

Bump up the version of the HASH_OBJECT's header to '3' so that previously
written state can be resumed if an event sequence object is encountered
and we only unmarshal an event sequence object when the version is at least
'3'.

Fixes issue #259.

### DIFF
--- a/src/tpm2/NVMarshal.c
+++ b/src/tpm2/NVMarshal.c
@@ -2245,7 +2245,7 @@ HMAC_STATE_Unmarshal(HMAC_STATE *data, BYTE **buffer, INT32 *size)
 }
 
 #define HASH_OBJECT_MAGIC 0xb874fe38
-#define HASH_OBJECT_VERSION 2
+#define HASH_OBJECT_VERSION 3
 
 static UINT16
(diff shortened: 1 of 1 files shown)


diabonas (approved):
Thank you very much for your quick fix! I confirm that this resolves the issue.

stefanberger:
> Thank you very much for your quick fix! I confirm that this resolves the issue.

Did you find this by running a test suite or running it by command line?

diabonas:
> Did you find this by running a test suite or running it by command line?

I re-ran the minimal example from #259, which works fine after the change in this PR.

The first time I encountered the issue was when trying to run the tpm2-tools integration test suite with libtpms 0.8.5, which lead to a failure of the [`pcrevent.sh`](https://github.com/tpm2-software/tpm2-tools/blob/master/test/integration/tests/pcrevent.sh) test. This test failure is fixed now, so the tpm2-tools test suite passes (as well as the tpm2-tss, tpm2-abrmd and tpm2-pkcs11 test suites, which weren't affected by this in the first place).

PR #262, opened 2021-09-15
Fix build and runtime for OpenSSL 3.0

stefanberger:
This PR fixes various aspects of the build of libtpms for OpenSSL 3.0 and also some runtime aspects of it.


stefanberger:
I tested this now with OpenSSL 3.0 downloaded from koji for Fedora and also tested this on various systems with OpenSSL 1.1.x and the swtpm test suite (for TPM 2 via IBMTSS 2). All worked fine.

PR #267, opened 2021-09-24
Implement some of the functions in Helpers.c with OpenSSL 3.0 APIs (part 1)

stefanberger:
Implement some of the functions in Helpers.c with OpenSSL 3.0 APIs.

### DIFF
--- a/src/tpm2/crypto/openssl/Helpers.c
+++ b/src/tpm2/crypto/openssl/Helpers.c
@@ -408,20 +408,99 @@ ComputePrivateExponentD(
     return pOK;
 }
 
+/* Build an RSA key from the given BIGUMs. The caller must always free
+ * the passed BIGNUMs.
+ */
+static int
+BuildRSAKey(EVP_PKEY **ppkey, // OUT
+            const BIGNUM *N, const BIGNUM *E, const BIGNUM *D,
+            const BIGNUM *P, const BIGNUM *Q,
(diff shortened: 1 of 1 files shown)


stefanberger:
@elmarco I think the abstractions here are good now for supporting OpenSSL 3.0 and the older versions via the new function BuildRSAKey(), but in terms of performance this needs a lot more memory allocations for duplicating BIGNUMs now (in the OpenSSL 1.x version of BuildRSAKey()) than before with only OpenSSL 1.x support. 

I think this is for 'after libtpms v0.9.0'.

elmarco:
@stefanberger I am not that familiar with openssl, but from a 10000ft view, looks ok. Why do you have those `while (1)`? (also in tree already). It should simply abort().

stefanberger:
> Why do you have those `while (1)`? (also in tree already). It should simply abort().

Those are only in test code so I can still hook a debugger onto it and inspect variables if that ends up being helpful at all...

elmarco:
> > Why do you have those `while (1)`? (also in tree already). It should simply abort().
> 
> Those are only in test code so I can still hook a debugger onto it and inspect variables if that ends up being helpful at all...

I see, well abort() should give you a similar result, with `coredumpctl debug`

stefanberger:
Closing this PR now and will merge PR #269 instead.

PR #270, opened 2021-10-09
Support SM3 and SM4

JerryDevis:
Support SM3 and SM4 for libtpms.


JerryDevis:
Your suggestion is indeed a good solution for now, I will modify the patch again.

stefanberger:
You can split it up in two PRs, first SM3 and then SM4.

For these to be compileable on Travis I think we'll have to upgrade to Ubuntu 20.04... but let's do that later.

Issue #215, opened 2021-05-17
OpenSSL 3.0 support


elmarco:
libtpms fails to build with openssl 3.0 (https://wiki.openssl.org/index.php/OpenSSL_3.0).

Porting the code to support 3.0 would help with FIPS compliancy.

However, since a lot of code is imported from the specification, afaik, how do we tackle this? How can we share the work?

stefanberger:
> How can we share the work?

I haven't tried it with 3.0 so far so I cannot say what it will take to port it. I have heared that the origin TPM 2 code works with OpenSSL 3.0 out of the box, so it's probably the FIPS compliancy that I persued for RSA, symmetric keys and ECC a while ago that breaks the build now.

stefanberger:
By the way, when configuring with `--disable-use-openssl-function` then this will build more or less the original TPM 2 code that uses by far less of the OpenSSL APIs.

elmarco:
Building with `-Wno-error` allows to see what needs to be updated:

```
tpm2/crypto/openssl/CryptCmac.c: In function ‘CryptCmacData’:
tpm2/crypto/openssl/CryptCmac.c:128:13: warning: ‘AES_set_encrypt_key’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  128 |             FOR_EACH_SYM(ENCRYPT_CASE)
      |             ^~~~~~~~~~~~
In file included from ./tpm2/crypto/openssl/TpmToOsslSym.h:72,
                 from ./tpm2/crypto/openssl/LibSupport.h:94,
                 from ./tpm2/Tpm.h:73,
                 from tpm2/crypto/openssl/CryptCmac.c:69:
/usr/include/openssl/aes.h:51:5: note: declared here
   51 | int AES_set_encrypt_key(const unsigned char *userKey, const int bits,
      |     ^~~~~~~~~~~~~~~~~~~
tpm2/crypto/openssl/CryptCmac.c:128:13: warning: ‘AES_encrypt’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  128 |             FOR_EACH_SYM(ENCRYPT_CASE)
      |             ^~~~~~~~~~~~
In file included from ./tpm2/crypto/openssl/TpmToOsslSym.h:72,
                 from ./tpm2/crypto/openssl/LibSupport.h:94,
                 from ./tpm2/Tpm.h:73,
[...]

PR #271, opened 2021-10-11
Support SM3 and SM4

JerryDevis:
we add a compiling macro(i.e. --with-openssl-smX) to control SM3 and SM4, default not support.


JerryDevis:
We do need to add a CI machine that supports SM3/SM4, like ubuntu 20.04 as you mentioned.

src/tpm2/crypto/openssl/Helpers.c
```diff
@@ -623,3 +628,28 @@ OpenSSLCryptRsaGenerateKey(
...
+    uint32_t len = SM3_256_DIGEST_SIZE;
+    int ret = EVP_DigestFinal_ex(*c, md, &len);
+    if (ret != SM3_SUCCESS || len != SM3_256_DIGEST_SIZE) {
+        ret = SM3_FAIL;
+    }
+    EVP_MD_CTX_destroy(*c);
+    *c = NULL;
+    return ret;
+}
+#endif
```
stefanberger, line 709: new line after #endif

src/tpm2/crypto/openssl/Helpers.c
```diff
@@ -623,3 +628,28 @@ OpenSSLCryptRsaGenerateKey(
...
+#if ALG_SM3_256
+int sm3_init(SM3_TPM_CTX *c)
+{
+    *c = EVP_MD_CTX_new();
+    if (*c == NULL) {
+        return SM3_FAIL;
+    }
+    return EVP_DigestInit_ex(*c, EVP_sm3(), NULL);
+}
+int sm3_update(SM3_TPM_CTX *c, const void *data, size_t len)
```
stefanberger, line 694: please add an empty line between functions

src/tpm2/crypto/openssl/Helpers.c
```diff
@@ -623,3 +628,28 @@ OpenSSLCryptRsaGenerateKey(
...
+{
+    return EVP_DigestUpdate(*c, data, len);
+}
+int sm3_final(unsigned char *md, SM3_TPM_CTX *c)
+{
+    uint32_t len = SM3_256_DIGEST_SIZE;
+    int ret = EVP_DigestFinal_ex(*c, md, &len);
+    if (ret != SM3_SUCCESS || len != SM3_256_DIGEST_SIZE) {
```
stefanberger, line 702: empty line after variables declaration.

stefanberger: Please just use '1' for OpenSSL function success checking, no SM3_SUCCESS.

src/tpm2/NVMarshal.c
```diff
@@ -2026,7 +2026,145 @@ tpmHashStateSHA512_Unmarshal(SHA512_CTX *data, BYTE **buffer, INT32 *size,
...
+static UINT16
+tpmHashStateSM3_256_Unmarshal(tpmHashStateSM3_256_t *data, BYTE **buffer, INT32 *size,
+                             UINT16 hashAlg)
+{
+    UINT16 rc = TPM_RC_SUCCESS;
+    UINT16 array_size;
+    NV_HEADER hdr;
+    SM3_CTX *sm3_ctx = NULL;
+    
+    (*data) = EVP_MD_CTX_new();
```
stefanberger, line 2092: I think we shouldn't create a context here but fill the data just like we fill it with sha1 etc.

stefanberger: The risk with sm3 is that its data structure has always been opaque it seems. That leaves OpenSSL the opportunity to shuffle things around. They may not do that but the risk is there. In contrast to that we have the SHAT_CTX, SHA256_CTX, etc. that are available via sha.h, so there's less of a risk with them on that level.

src/tpm2/crypto/openssl/Helpers.c
```diff
@@ -623,3 +628,28 @@ OpenSSLCryptRsaGenerateKey(
...
+        ret = SM3_FAIL;
+    }
+    EVP_MD_CTX_destroy(*c);
+    *c = NULL;
+    return ret;
```
stefanberger, line 707: return SM3_SUCCESS

src/tpm2/crypto/openssl/Sm3Helper_fp.h
```diff
@@ -0,0 +1,25 @@
...
+   SM3_WORD A, B, C, D, E, F, G, H;
+   SM3_WORD Nl, Nh;
+   SM3_WORD data[SM3_LBLOCK];
+   unsigned int num;
+} SM3_CTX;
+int sm3_init(SM3_TPM_CTX *c);
+int sm3_update(SM3_TPM_CTX *c, const void *data, size_t len);
+int sm3_final(unsigned char *md, SM3_TPM_CTX *c);
+#endif
+#endif
```
stefanberger, line 25: Large parts of this file are already in TpmToOsslHash.h and shouldn't be duplicated. Please try to use that file.

src/tpm2/crypto/openssl/TpmToOsslHash.h
```diff
@@ -75,25 +75,9 @@
...
-    unsigned int num;
-} SM3_CTX;
-
-int sm3_init(SM3_CTX *c);
-int sm3_update(SM3_CTX *c, const void *data, size_t len);
-int sm3_final(unsigned char *md, SM3_CTX *c);
-#   endif // OpenSSL < 1.2
+    #include "Sm3Helper_fp.h"
+#endif
 #endif // ALG_SM3_256
```
stefanberger, line 81: Why do you remove it from here? I'd rather not move this here to keep the differences to the reference code at a minimum.

src/tpm2/crypto/openssl/TpmToOsslHash.h
```diff
@@ -108,10 +92,7 @@ int sm3_final(unsigned char *md, SM3_CTX *c);
 #define tpmHashStateSHA256_t      SHA256_CTX
 #define tpmHashStateSHA384_t      SHA512_CTX
 #define tpmHashStateSHA512_t      SHA512_CTX
-#define tpmHashStateSM3_256_t     SM3_CTX
-#if ALG_SM3_256
-#   error "The version of OpenSSL used by this code does not support SM3"
-#endif
+#define tpmHashStateSM3_256_t     SM3_TPM_CTX
 /*     The defines below are only needed when compiling CryptHash.c or CryptSmac.c. This isolation
```
stefanberger, line 96: Also this should stay here.

src/tpm2/NVMarshal.c
stefanberger: I know... but it's not going to be that simple. See my comment over there: https://github.com/stefanberger/libtpms/issues/215#issuecomment-842757244

PR #292, opened 2022-01-20
Support ecc encrypt

JerryDevis:


src/tpm2/AlgorithmTests.c
```diff
@@ -628,6 +628,54 @@ LoadEccPoint(
...
+
+    CLEAR_BOTH(scheme);
+    EccKeyInitialize(&testObject);
+    eccScheme.scheme = scheme;
+    eccScheme.details.mgf1.hashAlg = DEFAULT_TEST_HASH;
+    memcpy(testInput.t.buffer, c_ecTestPlaintext, sizeof(c_ecTestPlaintext));
+    testInput.t.size = sizeof(c_ecTestPlaintext);
+    if(TPM_RC_SUCCESS != CryptEccEncrypt(&testObject, &eccScheme,
+                    &testInput, &C1, &C2, &C3)) {
+    SELF_TEST_FAILURE;
```
stefanberger, line 667: indentation is incorrect

src/tpm2/TpmProfile.h
```diff
@@ -563,10 +563,10 @@
 #define CC_Duplicate                        CC_YES
 #endif
 #ifndef CC_ECC_Decrypt
-#define CC_ECC_Decrypt                      (CC_NO && ALG_ECC)
+#define CC_ECC_Decrypt                      (CC_YES && ALG_ECC)
 #endif
 #ifndef CC_ECC_Encrypt
-#define CC_ECC_Encrypt                      (CC_NO && ALG_ECC)
+#define CC_ECC_Encrypt                      (CC_YES && ALG_ECC)
```
stefanberger, line 569: Please leave it disabled for now. I may take the code but not sure when I will enable it.

src/tpm2/crypto/CryptEccCrypt_fp.h
```diff
@@ -62,6 +62,32 @@
...
+    OBJECT                  *key,           // IN: key used for data recovery
+    TPMT_KDF_SCHEME         *scheme,        // IN: scheme to use.
+    TPM2B_MAX_BUFFER        *plainText,     // OUT: the recovered text
+    TPMS_ECC_POINT          *c1,            // IN: public ephemeral key
+    TPM2B_MAX_BUFFER        *c2,            // IN: obfuscated text
+    TPM2B_DIGEST            *c3             // IN: digest of ephemeral key
+                                            //      and plainText
+);
+
+#endif
```
stefanberger, line 93: missing newline

tests/nvram_offsets.c
```diff
@@ -21,7 +21,7 @@ int main(void)
     }
 
     /* Check size of auditCommands that expands with new commands */
-#define PD_AUDIT_COMMANDS_EXP_SIZE 14
+#define PD_AUDIT_COMMANDS_EXP_SIZE 15
```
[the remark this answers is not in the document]
JerryDevis: OK, will ECC encryption and decryption command be considered to support in the future?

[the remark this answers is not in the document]
stefanberger: We can consider it. Problem by now is that every additional enabled command that alters the format of the state breaks migrateability to older versions of libtpms, which breaks certain scenarios of VM migration. However, due to requests to support runtime-disabled algorithms we *may* have to alter the format of the TPM's state because of that and the ECC encryption stuff could go in with it.

PR #294, opened 2022-02-19
tpm2: Do not call EVP_PKEY_CTX_set0_rsa_oaep_label() for label of size 0 (OSSL 3)

JuergenReppSIT:
Openssl 3.0 did return an error if EVP_PKEY_CTX_set0_rsa_oaep_label was called
with label size 0. The function should only be called if the size of the label
is greater 0.
With this fix TPM2_RSA_Encrypt/Decrypt did work with OpenSSL 1.1 and 3.0
for encryption without label.

Signed-off-by: Juergen Repp <juergen.repp@sit.fraunhofer.de>

### DIFF
--- a/src/tpm2/crypto/openssl/CryptRsa.c
+++ b/src/tpm2/crypto/openssl/CryptRsa.c
@@ -1356,10 +1356,9 @@ CryptRsaEncrypt(
                 if (tmp == NULL)
                     ERROR_RETURN(TPM_RC_FAILURE);
                 memcpy(tmp, label->buffer, label->size);
+                if (EVP_PKEY_CTX_set0_rsa_oaep_label(ctx, tmp, label->size) <= 0)
+                    ERROR_RETURN(TPM_RC_FAILURE);
             }
-            // label->size == 0 is supported
-            if (EVP_PKEY_CTX_set0_rsa_oaep_label(ctx, tmp, label->size) <= 0)
-                ERROR_RETURN(TPM_RC_FAILURE);
             tmp = NULL;
             break;
           default:


src/tpm2/crypto/openssl/CryptRsa.c
```diff
@@ -1444,7 +1443,7 @@ CryptRsaDecrypt(
                     ERROR_RETURN(TPM_RC_FAILURE);
                 memcpy(tmp, label->buffer, label->size);
 
-                if (EVP_PKEY_CTX_set0_rsa_oaep_label(ctx, tmp, label->size) <= 0)
```
stefanberger, line 1447: I don't think in this case it is necessary to again check label->size > 0. The whole block is already protected with this check.

src/tpm2/crypto/openssl/CryptRsa.c
```diff
@@ -1357,8 +1357,7 @@ CryptRsaEncrypt(
                     ERROR_RETURN(TPM_RC_FAILURE);
                 memcpy(tmp, label->buffer, label->size);
             }
-            // label->size == 0 is supported
-            if (EVP_PKEY_CTX_set0_rsa_oaep_label(ctx, tmp, label->size) <= 0)
+            if (label->size > 0 && EVP_PKEY_CTX_set0_rsa_oaep_label(ctx, tmp, label->size) <= 0)
```
stefanberger, line 1360: Can you move the EVP_PKEY_CTX_set0_rsa_oaep_label() into the block above it that already checks for label->size > 0.

stefanberger (changes requested):

stefanberger:
FYI: This here is the OpenSSL 1.1.1 that still allowed label to be NULL: https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/rsa/rsa_pmeth.c#L545-L550

stefanberger:
Can you also please adjust the title of the patch to something like `tpm2: Do not call EVP_PKEY_CTX_set0_rsa_oaep_label() for label of size 0 (OSSL 3)`

src/tpm2/crypto/openssl/CryptRsa.c
```diff
@@ -1444,7 +1443,7 @@ CryptRsaDecrypt(
                     ERROR_RETURN(TPM_RC_FAILURE);
                 memcpy(tmp, label->buffer, label->size);
 
-                if (EVP_PKEY_CTX_set0_rsa_oaep_label(ctx, tmp, label->size) <= 0)
+                if (label->size > 0 && EVP_PKEY_CTX_set0_rsa_oaep_label(ctx, tmp, label->size) <= 0)
```
stefanberger, line 1446: I don't think in this case it is necessary to again check label->size > 0. The whole block is already protected with this check.

JuergenReppSIT:
> FYI: This here is the OpenSSL 1.1.1 that still allowed label to be NULL: https://github.com/openssl/openssl/blob/OpenSSL_1_1_1-stable/crypto/rsa/rsa_pmeth.c#L545-L550
@stefanberger Thank you for the info. Yes in our CI for tss we have three test cases with label size 0 which did always work. During the first tests with Ubuntu 22.04 using libtpms installed via apt the errors did occur.

stefanberger:
Thanks!

stefanberger:
@JuergenReppSIT More recent TPM 2 code is under the BSD-2-Clause license. Do you mind this contribution to move under this 'new' license once the file that you applied the patch to moves under that license? If not I would have to revert the patch...

JuergenReppSIT:
> @JuergenReppSIT More recent TPM 2 code is under the BSD-2-Clause license. Do you mind this contribution to move under this 'new' license once the file that you applied the patch to moves under that license? If not I would have to revert the patch...

@stefanberger Yes, I would agree to that.

PR #305, opened 2022-03-30
Fix a few uninitialized variables

cpaelzer:
Hi,
while collecting some stable fixes for Ubuntu I found one patch that - AFAICS - wasn't upstreamed yet.
But I think these changes make sense to the project in general just as much.
Therefore I've prepared those for your consideration.

I kept Authorship info intact (= @vorlonofportland) but signed-off myself as I reworked them to match your commit styles.

P.S. I haven't found these as issues or open PRs nor in any recent branches. If I just missed them I beg your pardon and feel free to close.

### DIFF
--- a/src/tpm12/tpm_nvram.c
+++ b/src/tpm12/tpm_nvram.c
@@ -1290,7 +1290,7 @@ TPM_RESULT TPM_Process_NVReadValue(tpm_state_t *tpm_state,
     TPM_BOOL			physicalPresence;
     TPM_BOOL			isGPIO = FALSE;
     BYTE 			*gpioData = NULL;
-    TPM_NV_DATA_SENSITIVE	*d1NvdataSensitive;
+    TPM_NV_DATA_SENSITIVE	*d1NvdataSensitive = NULL;
     uint32_t			s1Last;
     
     /* output parameters  */
--- a/src/tpm2/Marshal.c
+++ b/src/tpm2/Marshal.c
@@ -2203,7 +2203,7 @@ UINT16
 TPM2B_NV_PUBLIC_Marshal(TPM2B_NV_PUBLIC *source, BYTE **buffer, INT32 *size)
 {
     UINT16 written = 0;
-    BYTE *sizePtr;
+    BYTE *sizePtr = NULL;
 
     if (buffer != NULL) {
 	sizePtr = *buffer;


src/tpm2/Marshal.c
```diff
@@ -2203,7 +2203,7 @@ UINT16
 TPM2B_NV_PUBLIC_Marshal(TPM2B_NV_PUBLIC *source, BYTE **buffer, INT32 *size)
 {
     UINT16 written = 0;
-    BYTE *sizePtr;
+    BYTE *sizePtr = NULL;
```
stefanberger, line 2206: How did you find this?

sizePtr is initialized when buffer != NULL and then later on access also again when buffer != NULL.

stefanberger: Oh, gcc on s390x.. uuhhh

src/tpm12/tpm_nvram.c
```diff
@@ -1290,7 +1290,7 @@ TPM_RESULT TPM_Process_NVReadValue(tpm_state_t *tpm_state,
     TPM_BOOL			physicalPresence;
     TPM_BOOL			isGPIO = FALSE;
     BYTE 			*gpioData = NULL;
-    TPM_NV_DATA_SENSITIVE	*d1NvdataSensitive;
+    TPM_NV_DATA_SENSITIVE	*d1NvdataSensitive = NULL;
```
stefanberger, line 1293: From looking at `TPM_NVIndexEntries_GetEntry()`, which would return an error code in case it could not initialize d1NvdataSensitive (found = FALSE) and otherwise have it initialized, this seems to be a false positive.

stefanberger:
Thanks for the PR. I would say both of these are false positives and I think we should mark them as such in the patch title. The title of the patches should probably be: Initialize variables due to gcc complaint (s390x, false positive)

src/tpm2/Marshal.c
cpaelzer: yeah both are gcc on s390x findings.

src/tpm12/tpm_nvram.c
cpaelzer: A false positive nowadays, but only if analyzed down to the level you did and only as long as `TPM_NVIndexEntries_GetEntry` or other potential future users are not changing the way they work.
Is there any drawback in initializing them explicitly?

cpaelzer:
> Thanks for the PR. I would say both of these are false positives and I think we should mark them as such in the patch title. The title of the patches should probably be: Initialize variables due to gcc complaint (s390x, false positive)

Done and pushed to my proposed branch.

Thanks for the quick review!

src/tpm12/tpm_nvram.c
stefanberger: > Is there any drawback in initializing them explicitly?

I don't mind these patches but I also have to determine whether they are false positives or real issues because in the end a NULL pointer dereference isn't much better.

src/tpm2/Marshal.c
stefanberger: I had these types of findings before also... quite a few of them on different architectures and most (if not all) of them were false positives.

cpaelzer: thanks for the explanation, I've already updated the branch an hour ago, should be good now.

src/tpm12/tpm_nvram.c
cpaelzer: yeah, thanks for sharing your thoughts

stefanberger:
@cpaelzer Do you still need to apply these 2 patches? I am asking because I don't need to apply them when building for my Ubuntu PPA for Bionic, Focal, and Jammy. So maybe gcc for s390x has been improved in the meantime.

cpaelzer:
> @cpaelzer Do you still need to apply these 2 patches? I am asking because I don't need to apply them when building for my Ubuntu PPA for Bionic, Focal, and Jammy. So maybe gcc for s390x has been improved in the meantime.

Hi @stefanberger, indeed I haven't checked that before.

If I read the history correctly the bad gcc was at the late days of Impish which is still out there and makes it worth to keep the changes merged either way.

But on your hint I was dropping it from my 0.9.3 merge and rebuilt on all platforms - indeed for Jammy I was able to drop it without consequence.

stefanberger:
The question is whether to backport them to 0.9.x to allow it to build without having to apply these patches. I guess you would want me to apply them to stable-0.9.

Issue #317, opened 2022-04-08
Compilation fails on OSS-Fuzz


elmarco:
A new compilation error appeared:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=46483

Step #12 - "compile-honggfuzz-address-x86_64": In file included from /src/libtpms/src/tpm_debug.c:42:
Step #12 - "compile-honggfuzz-address-x86_64": /src/libtpms/src/tpm_debug.h:69:9: error: 'printf' macro redefined [-Werror,-Wmacro-redefined]
Step #12 - "compile-honggfuzz-address-x86_64": #define printf(...) TPMLIB_LogPrintf(__VA_ARGS__);
Step #12 - "compile-honggfuzz-address-x86_64":         ^
Step #12 - "compile-honggfuzz-address-x86_64": /usr/include/x86_64-linux-gnu/bits/stdio2.h:110:11: note: previous definition is here
Step #12 - "compile-honggfuzz-address-x86_64": #  define printf(...) \
Step #12 - "compile-honggfuzz-address-x86_64":           ^
Step #12 - "compile-honggfuzz-address-x86_64": 1 error generated.
Step #12 - "compile-honggfuzz-address-x86_64": make[2]: *** [Makefile:2018: libtpms_la-tpm_debug.lo] Error 1
Step #12 - "compile-honggfuzz-address-x86_64": make[2]: *** Waiting for unfinished jobs....

It seems like a bad idea to "#define printf". Any chance we can replace all `printf()` occurences with `TPMLIB_LogPrintf()` instead? Or is the code in src/tpm12 still imported regularly from outside?

stefanberger:
PR #319 resolved the issue: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=46483#c1

PR #318, opened 2022-04-08
Remove #define printf

elmarco:
Mass replace printf() occurences with TPMLIB_LogPrintf().

Fixes: #317

Related to:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=46483

### DIFF
--- a/src/tpm12/tpm_admin.c
+++ b/src/tpm12/tpm_admin.c
@@ -96,7 +96,7 @@ TPM_RESULT TPM_LimitedSelfTestCommon(void)
     uint32_t	tv_sec;
     uint32_t	tv_usec;
 
-    printf(" TPM_LimitedSelfTestCommon:\n");
+    TPMLIB_LogPrintf(" TPM_LimitedSelfTestCommon:\n");
 #if 0
     if (rc == 0) {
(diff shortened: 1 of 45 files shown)


stefanberger:
Thanks for the PR. I wonder whether a backport-able easy patch wouldn't just be

```
#ifdef printf
#undef printf
#endif
```

I would apply this one first before yours...

elmarco:
As long as nothing is using the actual system printf(), that should work. But that could have surprise consequences. I think we should really not #redefine system/lib macros...

stefanberger:
> As long as nothing is using the actual system printf(), that should work. But that could have surprise consequences. I think we should really not #redefine system/lib macros...

Correct. 

My logic here is that we had #define'd printf now for such a long time that this seems to be the simplest fix.

elmarco:
> My logic here is that we had #define'd printf now for such a long time that this seems to be the simplest fix.

My point is that this is not really a fix if you still replace the system printf(). OSS-Fuzz might be redirecting it to some extra checks, fuzzing hint, services etc.

stefanberger:
If you want this to be merged, you would have to rebase the series.

stefanberger:
Series seems stale.

PR #322, opened 2022-05-20
nvfile: Free allocated memory on failure

rosslagerwall:
In TPM_NVRAM_LoadData(), there is an unlikely path where the function
will return an error code but still expect the caller to free the
allocated data. At least some of the callers don't handle this correctly
so ensure that the caller only needs to free data if the function
returns success.

Reported by Coverity.

Signed-off-by: Ross Lagerwall <ross.lagerwall@citrix.com>

### DIFF
--- a/src/tpm_nvfile.c
+++ b/src/tpm_nvfile.c
@@ -267,6 +267,11 @@ TPM_RESULT TPM_NVRAM_LoadData(unsigned char **data,     /* freed by caller */
             printf(" TPM_NVRAM_LoadData: Closed file %s\n", filename);
         }
     }
+
+    if (rc) {
+        free(*data);
+        *data = NULL;
+    }
     return rc;
 }
 


src/tpm_nvfile.c
```diff
@@ -262,6 +264,8 @@ TPM_RESULT TPM_NVRAM_LoadData(unsigned char **data,     /* freed by caller */
         if (irc != 0) {
             printf("TPM_NVRAM_LoadData: Error (fatal) closing file %s\n", filename);
             rc = TPM_FAIL;
+            free(*data);
+            *data = NULL;
```
stefanberger, line 268: Can you move this to the bottom:

```
if (rc && data != NULL) {
    free(*data);
    *data = NULL;
}
return rc;
```

stefanberger:
Thanks for the PR.
This is in a practically unused code path. At least swtpm doesn't use it...

src/tpm_nvfile.c
rosslagerwall: Updated. I skipped the `data != NULL` check since `data` is dereferenced at the beginning of the function which implies it is non-NULL.

PR #323, opened 2022-05-23
tpm2: Fix size check in CryptSecretDecrypt

rosslagerwall:
Check the secret size against the size of the buffer, not the size
member that has not been set yet.

Reported by Coverity.

Signed-off-by: Ross Lagerwall <ross.lagerwall@citrix.com>

### DIFF
--- a/src/tpm2/CryptUtil.c
+++ b/src/tpm2/CryptUtil.c
@@ -732,7 +732,7 @@ CryptSecretDecrypt(
 					     nonceCaller->t.size);
 			      }
 			  // make sure secret will fit
-			  if(secret->t.size > data->t.size)
+			  if(secret->t.size > sizeof(data->t.buffer))
 			      return TPM_RC_FAILURE;
 			  data->t.size = secret->t.size;
 			  // CFB decrypt, using nonceCaller as iv


stefanberger:
Thanks.

Kenneth Goldman reported it to TCG. I'll merge your patch and will backport it once I hear back.

stefanberger:
I am also running the code through Coverity but I didn't see this one reported. How was it reported to you?

stefanberger:
In my version of Coverity (scan.coverity.com) this patch did not resolve any previously reported issues. Are you using the same version of Coverity?

rosslagerwall:
At Citrix, we're using our own internal instance of Coverity (2021.12.1), not the scan.coverity one for OSS projects. I used the following command-line to cov-analyze:

```
cov-analyze --dir coverity --strip-path /builddir/build/BUILD/libtpms-0.9.4 --cpp --aggressiveness-level high --all --rule --disable-parse-warnings --enable-fnptr --disable-fb -j auto $MODEL_ARGS $*
```

Maybe it is the agressiveness level that matters? It generates hundreds of issues, most of which appear to be false positives at first glance. At least some of them are valid though.

PR #340, opened 2022-07-12
update libressl support

williamcroberts:
Revert the patch checking if LIBRESSL is defined and include opensslv.h so that way openssl version number macros work as expected.


williamcroberts:
This is tested against LibreSSL 3.5.3 and 2.7.2

src/tpm12/tpm_crypto.c
```diff
@@ -133,7 +134,7 @@ static TPM_RESULT TPM_SymmetricKeyData_Crypt(unsigned char *data_out,
 
 #include <openssl/aes.h>
 
-#if defined(__OpenBSD__) || defined(LIBRESSL_VERSION_NUMBER)
+#if defined(__OpenBSD__)
```
stefanberger, line 137: You can even remove the `#if` branch and the include of opensslv.h doesn't seem to be necessary (anymore?) on the version of OpenBSD where I just test compiled this.

stefanberger: I only ever test compile this with one version of OpenBSD which happens to have LibreSSL 3.3.2 ...

williamcroberts:
> This is tested against LibreSSL 3.5.3 and 2.7.2

Ahh LibreSSL is not really a drop in replacement to OpenSSL and it seems that they jacked OpenSSL Version number to  0x20000000L. This thread is of interest, https://github.com/libressl-portable/portable/issues/573, it looks like proper code should be checking for features in configure or using libressl version number which is going to suck as well.

Let's just revert that one patch....

src/tpm12/tpm_crypto.c
williamcroberts: > Who is interested in supporting LibreSSL ?

I just need to build it to test tpm2-tss against, its generally easier to just have everything on the system compile against one version so you don't accidentally bork some env setting and end up compiling/linking against things you didn't expect.

> If it's not just a simple drop-in replacement for OpenSSL someone will have to sign up for maintainership of LibreSSL support and add the name to this file here along with 'maintainer of LibreSSL support': https://github.com/stefanberger/libtpms/blob/master/MAINTAINERS

Yeah I see that, I see that they essentially are a 1.1.1 with some missing functionality and things getting pulled in. While libtpms will work against pretty ancient versions of libressl, swtpm needs the oaep support in version libressl-v3.1.0 or higher.

So this patch works, as intended, because they set OPENSSL_VERSION_NUMBER 0x020000000L which correctly puts us on the "new 1.1.1" API. However, when libtpms goes to 3.0, this logic will need to be revamped. We could also likely drop the BSD check and just check for the version number between OPENSSL and LIBRESSL.
```C
#if defined(__OpenBSD__)
 # define OPENSSL_OLD_API
#elif OPENSSL_VERSION_NUMBER < 0x10100000
 #define OPENSSL_OLD_API
#endif
```

PR #343, opened 2022-07-15
Add support for setting of profiles for a TPM 2

stefanberger:
This PR adds support for setting of a profile for a TPM 2 that restricts what algorithms and commands are available to the TPM 2. With profiles it will be possible to provide forwards and backwards compatibility when upgrading or downgrading libtpms. A TPM 2 instance that was created with libtpms v0.10 can be upgrade to a future/later version of libtpms but then will also allow for downgrading to libtpms back to v0.10. To achieve this, the TPM 2 instance should not gain more functionality when it is running with a future/later version of libtpms so that applications using the TPM 2 (from inside a VM) do not loose TPM 2 functionality when libtpms is downgraded again. One caveat here is that this assumes that new functionality is gained through new commands ( that are not enable for this TPM 2 instance ) implemented by the future/later libtpms rather than extensions to existing commands ( which would complicate the code to enable / disable certain parts of an existing commands ).

This PR introduces a new API call `TPMLIB_SetProfile()` that allows a user (swtpm) to set a TPM 2 RuntimeProfile when the TPM 2 is first started and no TPM 2 state file exists. This TPM 2 profile will be firmly associated with the TPM 2 instance and cannot be changed anymore. Therefore, subsequent calls to the this API will not have any effect on the profile that was previously set for a TPM 2 instance.  A profile is a JSON object formatted as string that must contain the name of a profile known to libtpms and may contain the set of algorithms to enable for the TPM 2 instance. Algorithms are described via a comma-separated list of verbs known to libtpms. The following is are example profiles that can be set:

```
    {"name":"default"}

    {
        "name":"default",
        "algorithms":"rsa,rsa-min-size=1024,tdes-min-size=128,hmac,aes,\
                      aes-min-size=128,mgf1,keyedhash,xor,sha256,sha384,\
                      sha512,null,rsassa,rsaes,rsapss,oaep,ecdsa,ecdh,ecdaa,\
                      sm2,ecschnorr,ecmqv,kdf1-sp800-56a,kdf2,kdf1-sp800-108,\
                      ecc,ecc-min-size=192,symcipher,camellia,\
                      camellia-min-size=128,cmac,ctr,ofb,cbc,cfb,ecb"
    }
```

The first profile enables the current default profile. The second profile enables the current default profile and explicitly restricts the algorithms that the TPM 2 offers to applications with the given list.

The user does not have control over the set of commands that a profile enables but only algorithms (as shown above).

A TPM 2 instances that was previously created under libtpms v0.9 can be upgraded to libtpms v.10 or later and then be downgraded again to libtpms v0.9. This TPM 2 instance does not carry a profile as part of its state but runs with the 'default' profile that provides access to the commands and algorithms of libtpms v0.9 when it is started with libtpms v0.10 or a later version.

If an instance created by a future/later version of libtpms is read, then the reading of the state may fail if the 'stateFormatLevel' field contained in the profile (read from the state) holds a value greater than the currently supported one.


stefanberger:
What should the name if the FIPS profile be that disables various algorithms that OpenSSL disables on a host if FIPS mode is enabled? Should it be just 'FIPS' or 'FIPS-140' or 'OSSL-FIPS'? 

FIPS 140-2 from 10/2021 still mentions SHA-1 and doesn't seem to say it's deprecated: https://csrc.nist.gov/CSRC/media/Publications/fips/140/2/final/documents/fips1402annexa.pdf

stefanberger:
The current code allows to migrate TPM state initially created by libtpms v0.9.5 to v0.10 and back again.

However, if the state is created with v0.10 and an old swtpm version or without specifying the NULL profile using swtpm_setup, then migration to v0.9.x fails because the profile of v0.10 is used if no profile is selected.

The following has to be run to get v0.10 to be compatible with v0.9.x:
```
swtpm_setup --tpmstate /var/lib/libvirt/swtpm/ecc221c4-6bb9-423f-ac31-72244fdbb1a1/tpm2/ --create-ek-cert --create-platform-cert --overwrite --profile '{"name":"null"}'
```

src/tpm2/RuntimeAlgorithm.c
```diff
@@ -0,0 +1,506 @@
...
+#include "GpMacros.h"
+#include "tpm_library_intern.h"
+
+#define ALGO_SEPARATOR_C ','
+#define ALGO_SEPARATOR_STR ","
+
+struct KeySizes {
+    BOOL enabled;
+    UINT16 size;
+};
```
stefanberger, line 57: Add a column `stateFormatLevel` that indicates which stateFormatLevel is required to write a key of this size?

src/tpm2/RuntimeAlgorithm.c
```diff
@@ -0,0 +1,506 @@
...
+    { .enabled = false        , .size = 0 },
+};
+
+static const struct {
+    const char   *name;
+    union {
+	const struct KeySizes *keySizes;
+    } u;
+    BOOL          canBeDisabled;
+} s_AlgorithmProperties[NUM_ENTRIES_ALGORITHM_PROPERTIES] = {
```
stefanberger, line 105: Add a column `stateFormatLevel` that indicates which stateFormatLevel is required to write this algorithm?

src/tpm2/RuntimeCommands.c
```diff
@@ -0,0 +1,363 @@
...
+#include "tpm_library_intern.h"
+
+#if VENDOR_COMMAND_COUNT != 0
+# error Vendor commands not supported
+#endif
+
+
+/* List of supported commands sorted by command codes */
+static const struct {
+    const char *name;
```
stefanberger, line 54: Add a column `stateFormatLevel` that indicates which stateFormatLevel is required when a certain command is supported?

src/tpm2/RuntimeProfile.c
```diff
@@ -0,0 +1,757 @@
...
+    const char *name;
+    const char *commandsProfile;
+    const char *algorithmsProfile;
+    /* StateFormatLevel drives the format the TPM's state is written in and
+     * how it is read.
+     * Once a version of libtpms is released this field must never change afterwards
+     * so that backwards compatibility for reading the state can be maintained.
+     * This basically locks the name of the profile to the stateFormatLevel.
+     */
+    unsigned int stateFormatLevel;
```
stefanberger, line 62: Can the stateFormatLevel be determined from the keysizes, types of algorithms, and commands enabled in a profile and be be removed from here? Will increasing the stateFormatLevel always be driven by new keysize, new algorithms, or new commands being enabled? (How to predict the future??)

src/tpm2/RuntimeAlgorithm.c
```diff
@@ -0,0 +1,762 @@
...
+    [TPM_ALG_SM2] = SIGNING(ALG_SM2, "sm2", false, 1),
+    [TPM_ALG_ECSCHNORR] = SIGNING(ALG_ECSCHNORR, "ecschnorr", false, 1),
+    [TPM_ALG_ECMQV] = OTHER(ALG_ECMQV, "ecmqv", true, 1),
+    [TPM_ALG_KDF1_SP800_56A] = HASH(ALG_KDF1_SP800_56A, "kdf1-sp800-56a", false, 1),
+    [TPM_ALG_KDF2] = HASH(ALG_KDF2, "kdf2", false, 1),
+    [TPM_ALG_KDF1_SP800_108] = HASH(ALG_KDF1_SP800_108, "kdf1-sp800-108", false, 1),
+    [TPM_ALG_ECC] = ASYMMETRIC(ALG_ECC, "ecc", s_KeySizesECC, false, 1),
+    [TPM_ALG_SYMCIPHER] = OTHER(ALG_SYMCIPHER, "symcipher", false, 1),
+    [TPM_ALG_CAMELLIA] = SYMMETRIC(ALG_CAMELLIA, "camellia", s_KeySizesCamellia, true, 1),
+    [TPM_ALG_SHA3_256] = HASH(ALG_SHA3_256, "sha3-256", false, 0), // not supported
```
stefanberger, line 150: this and the ones below must be 'true' even if not supported...

src/tpm2/RuntimeAlgorithm.c
```diff
@@ -0,0 +1,762 @@
...
+    [TPM_ALG_HMAC] = SIGNING(ALG_HMAC, "hmac", false, 1),
+    [TPM_ALG_AES] = SYMMETRIC(ALG_AES, "aes", s_KeySizesAES, false, 1), // never disable: context encryption
+    [TPM_ALG_MGF1] = HASH(ALG_MGF1, "mgf1", false, 1),
+    [TPM_ALG_KEYEDHASH] = HASH(ALG_KEYEDHASH, "keyedhash", false, 1),
+    [TPM_ALG_XOR] = OTHER(ALG_XOR, "xor", false, 1),
+    [TPM_ALG_SHA256] = HASH(ALG_SHA256, "sha256", false, 1),
+    [TPM_ALG_SHA384] = HASH(ALG_SHA384, "sha384", false, 1),
+    [TPM_ALG_SHA512] = HASH(ALG_SHA512, "sha512", true, 1),
+    [TPM_ALG_NULL] = OTHER(true, "null", false, 1),
+    [TPM_ALG_SM4] = SYMMETRIC(ALG_SM4, "sm4", s_KeySizesSM4, true, 0), // not supported
```
stefanberger, line 133: Must be 'true' even if not supported

src/tpm2/RuntimeAlgorithm.c
```diff
@@ -0,0 +1,762 @@
...
+    [TPM_ALG_RSASSA] = SIGNING(ALG_RSASSA, "rsassa", true, 1),
+    [TPM_ALG_RSAES] = ENCRYPTING(ALG_RSAES, "rsaes", true, 1),
+    [TPM_ALG_RSAPSS] = SIGNING(ALG_RSAPSS, "rsapss", true, 1),
+    [TPM_ALG_OAEP] = ENCRYPTING(ALG_OAEP, "oaep", false, 1), // never disable: CryptSecretEncrypt/Decrypt needs it
+    [TPM_ALG_ECDSA] = SIGNING(ALG_ECDSA, "ecdsa", false, 1),
+    [TPM_ALG_ECDH] = OTHER(ALG_ECDH, "ecdh", false, 1),
+    [TPM_ALG_ECDAA] = OTHER(ALG_ECDAA, "ecdaa", false, 1),
+    [TPM_ALG_SM2] = SIGNING(ALG_SM2, "sm2", false, 1),
```
stefanberger, line 141: should be able to disable this one

src/tpm_tpm2_interface.c
```diff
@@ -162,6 +163,7 @@ static TPM_RESULT TPM2_MainInit(void)
         NvCommit();
     }
 
+
```
stefanberger, line 166: Remove

man/man3/TPMLIB_GetInfo.pod
```diff
@@ -38,6 +38,131 @@ The following flags are defined and return JSON objects as shown:
...
+     "Description": "The profile enables the commands and algorithms that " \
+                    "were enabled in libtpms v0.9. This profile is " \
+                    "automatically used when the state does not have a " \
+                    "profile, for example when it was created by " \
+                    "libtpms v0.9 or before."
+   }
+ }
+
+The above profile is the 'null' profile, which, by its description allows to
+use the state written by libtpms with versions v0.9 and younger.
```
stefanberger, line 138: younger -> later

man/man3/TPMLIB_GetInfo.pod
```diff
@@ -38,6 +38,131 @@ The following flags are defined and return JSON objects as shown:
...
+
+=item B<TPMLIB_INFO_AVAILABLE_PROFILES> (since v0.10.0)
+
+This JSON object enumerates available profiles.
+
+ {
+   "AvailableProfiles": [
+     {
+       "Name": "default",
```
stefanberger, line 147: default -> default-v1

man/man3/TPMLIB_SetProfile.pod
```diff
@@ -0,0 +1,224 @@
...
+
+A profile is a JSON map in string format. It must contain the 'Name'
+field with the name of a known profile. It may contain a field
+'Algorithms' that holds a comma-separated list of algorithms (verbs) to
+enable or the field 'Commands' that holds a comma-separated list of ranges
+of command codes to enable.
+
+The following are examples of TPM 2 profiles:
+
+     {"Name":"default"}
```
stefanberger, line 56: default -> default-v1

PR #349, opened 2023-02-10
Replace deprecated symmetric functins with EVP functions

stefanberger:
This PR replaces most usage of old OpenSSL symmetric function with calls to EVP functions.


stefanberger:
A definite downside for using the EVP APIs is that they are computationally **much more** involved.

src/tpm2/crypto/openssl/TpmToOsslDesSupport.c
```diff
@@ -101,6 +102,7 @@ void TDES_encrypt(
 		     &ks[0], &ks[1], &ks[2],
 		     DES_ENCRYPT);
 }
+
```
stefanberger, line 105: stray  newline

PR #352, opened 2023-02-15
DO-NOT-MERGE: tpm2: Replace deprecated EC functions with EVP functions (OSSL 3)

stefanberger:
This PR replaces deprecated EC functions with EVP functions when compiled for OpenSSL 3.


stefanberger:
The changes work for the NIST curves but not for the bnp256 curve for example.

stefanberger:
Due to patched OpenSSL library in Fedora it is not possible to use the OpenSSL 3 API to set an ec group. See the issue described here: https://github.com/openssl/openssl/issues/20313#issuecomment-1434663610

This patch cannot be applied.

PR #358, opened 2023-02-28
tpm2: Check size of buffer before accessing it (CVE-2023-1017 & -1018)

stefanberger:
Check that there are sufficient bytes in the buffer before reading the cipherSize from it. Also, reduce the bufferSize variable by the number of bytes that make up the cipherSize to avoid reading and writing bytes beyond the buffer in subsequent steps that do in-place decryption.

This fixes CVE-2023-1017 & CVE-2023-1018.


elmarco:
lgtm,
Reviewed-by: Marc-André Lureau <marcandre.lureau@redhat.com>

PR #366, opened 2023-03-16
Add missing algorithm macro checks

ben-dav-lytle:
Ran into compile errors when disabling/enabling certain algorithms.

Signed-off-by: Ben Lytle <ben.dav.lytle@hpe.com>

### DIFF
--- a/src/tpm2/SymmetricTest.h
+++ b/src/tpm2/SymmetricTest.h
@@ -84,7 +84,7 @@ const SYMMETRIC_TEST_VECTOR   c_symTestValues[NUM_SYMS + 1] = {
       dataOut_AES128_CFB, dataOut_AES128_ECB}},
 #endif
 #if ALG_AES && AES_192
-    {TPM_ALG_AES_, 192, key_AES192, 16, sizeof(dataIn_AES192), dataIn_AES192,
+    {TPM_ALG_AES, 192, key_AES192, 16, sizeof(dataIn_AES192), dataIn_AES192,
      {dataOut_AES192_CTR, dataOut_AES192_OFB, dataOut_AES192_CBC,
       dataOut_AES192_CFB, dataOut_AES192_ECB}},
 #endif
--- a/src/tpm2/crypto/CryptSym.h
+++ b/src/tpm2/crypto/CryptSym.h
@@ -102,7 +102,10 @@
 //#define     TDES    DES[3]			/* libtpms commented */
 typedef union tpmCryptKeySchedule_t {
     FOR_EACH_SYM_WITHOUT_TDES(KEY_SCHEDULE)	/* libtpms changed from FOR_EACH_SYM */
+
+#if ALG_TDES                            // libtpms added
     tpmKeyScheduleTDES  tdes[3];		/* libtpms added */
+#endif                                  // libtpms added
 
 #if SYMMETRIC_ALIGNMENT == 8
     uint64_t            alignment;
--- a/src/tpm2/crypto/openssl/CryptEccSignature.c
+++ b/src/tpm2/crypto/openssl/CryptEccSignature.c
@@ -104,6 +104,7 @@ EcdsaDigest(
 /* e) n is the order of the curve */
 /* Error Returns Meaning */
 /* TPM_RC_NO_RESULT the result of the operation was zero or r (mod n) is zero */
+#if ALG_ECSCHNORR						// libtpms added
 static TPM_RC
 BnSchnorrSign(
 	      bigNum                   bnS,           // OUT: s component of the signature
@@ -130,6 +131,7 @@ BnSchnorrSign(
     BnDiv(NULL, bnS, bnT1, bnN);
     return (BnEqualZero(bnS)) ? TPM_RC_NO_RESULT : TPM_RC_SUCCESS;
 }
+#endif // ALG_ECSHNORR - libtpms added
 /* 10.2.12.3 Signing Functions */
 /* 10.2.12.3.1 BnSignEcdsa() */
 /* This function implements the ECDSA signing algorithm. The method is described in the comments
--- a/src/tpm2/crypto/openssl/Helpers.c
+++ b/src/tpm2/crypto/openssl/Helpers.c
@@ -91,6 +91,7 @@ BuildRSAKey(EVP_PKEY **ppkey, // OUT
 
 #if USE_OPENSSL_FUNCTIONS_SYMMETRIC
 
+#if ALG_TDES
 TPM_RC
 OpenSSLCryptGenerateKeyDes(
                            TPMT_SENSITIVE *sensitive    // OUT: sensitive area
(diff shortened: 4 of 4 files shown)


stefanberger:
> Ran into compile errors when disabling/enabling certain algorithms.

Can you describe how you ran into these issues?

src/tpm2/SymmetricTest.h
```diff
@@ -84,7 +84,7 @@ const SYMMETRIC_TEST_VECTOR   c_symTestValues[NUM_SYMS + 1] = {
       dataOut_AES128_CFB, dataOut_AES128_ECB}},
 #endif
 #if ALG_AES && AES_192
-    {TPM_ALG_AES_, 192, key_AES192, 16, sizeof(dataIn_AES192), dataIn_AES192,
+    {TPM_ALG_AES, 192, key_AES192, 16, sizeof(dataIn_AES192), dataIn_AES192,
```
stefanberger, line 87: Do you need AES_192? Iirc I left this in there on purpose since libtpms doesn't support it in its default config.

stefanberger:
> Ran into compile errors when disabling/enabling certain algorithms.

Ok giot i

stefanberger:
I did not mean to close the PR.

src/tpm2/crypto/CryptSym.h
```diff
@@ -102,7 +102,10 @@
 //#define     TDES    DES[3]			/* libtpms commented */
 typedef union tpmCryptKeySchedule_t {
     FOR_EACH_SYM_WITHOUT_TDES(KEY_SCHEDULE)	/* libtpms changed from FOR_EACH_SYM */
+
+#if ALG_TDES
```
stefanberger, line 106: Please append `// libtpms added`

src/tpm2/crypto/CryptSym.h
```diff
@@ -102,7 +102,10 @@
...
     tpmKeyScheduleTDES  tdes[3];		/* libtpms added */
+#endif
```
stefanberger, line 108: Please append `// libtpms added`

src/tpm2/crypto/openssl/CryptEccSignature.c
```diff
@@ -104,6 +104,7 @@ EcdsaDigest(
 /* e) n is the order of the curve */
 /* Error Returns Meaning */
 /* TPM_RC_NO_RESULT the result of the operation was zero or r (mod n) is zero */
+#if ALG_ECSCHNORR
```
stefanberger, line 107: Please append `// libtpms added`

src/tpm2/crypto/openssl/CryptEccSignature.c
```diff
@@ -130,6 +131,7 @@ BnSchnorrSign(
     BnDiv(NULL, bnS, bnT1, bnN);
     return (BnEqualZero(bnS)) ? TPM_RC_NO_RESULT : TPM_RC_SUCCESS;
 }
+#endif // ALG_ECSHNORR
```
stefanberger, line 134: Please append `// libtpms added`

ben-dav-lytle:
> > Ran into compile errors when disabling/enabling certain algorithms.
> 
> Can you describe how you ran into these issues?

Sorry, should've included that in PR description.

Specifically, I disabled `ALG_TDES` and `ALG_ECSHNORR` and got "undefined" compile errors where the new `#if` checks are.
I enabled AES_192, to test, and that was the only place that failed compilation.

I am testing different security profiles as far as supporting/not supporting specific algorithms; like FIPS 140-2 and 140-3, CNSA 1.0 and CNSA 2.0

stefanberger:
> I am testing different security profiles as far as supporting/not supporting specific algorithms; like FIPS 140-2 and 140-3, CNSA 1.0 and CNSA 2.0

These are TPM security profiles? The problem for libtpms is basically that we cannot really have compile-time options for security profiles for a TPM 2 but algorithms have to be made runtime-invisibel/inaccessible. See my PR #343 for those types of profiles.  The state of the TPM 2 and possibly downgrading requirements are a major challenge...

ben-dav-lytle:
> > I am testing different security profiles as far as supporting/not supporting specific algorithms; like FIPS 140-2 and 140-3, CNSA 1.0 and CNSA 2.0
> 
> These are TPM security profiles? The problem for libtpms is basically that we cannot really have compile-time options for security profiles for a TPM 2 but algorithms have to be made runtime-invisibel/inaccessible. See my PR #343 for those types of profiles. The state of the TPM 2 and possibly downgrading requirements are a major challenge...

Not TPM security profiles, but just conforming to FIPS or CNSA, as far as what algorithms are supported.

Yes, so similar to what you propose with #343; I was trying to take it one step further as completely compiling out the algorithms, but state migration would obviously not work. Is #343 still in testing?

I see you mention the profiles could only configure algorithms and not commands. Future plan for commands?

stefanberger:
> Yes, so similar to what you propose with #343; I was trying to take it one step further as completely compiling out the algorithms, but state migration would obviously not work. Is #343 still in testing?

I am deferring #343. Any later version of libtpms, whatever may be added to it, needs to be able to write the state of a previous version of libtpms driven by the applied profile. This can become a major headache including not being able to fix data in the state (which did happen before).

Compiling out algorithms is not really a choice because there will only be one version of libtpms in distros and it basically should enable all algorithms (and commands) and make algorithms and commands inaccessible if a certain profile is chosen. So it for example implements TDES but TDES is inaccessible & invisible (TPM2_GetCapability) because the profile doesn't support it.

> I see you mention the profiles could only configure algorithms and not commands. Future plan for commands?

Not currently...

ben-dav-lytle:
> Compiling out algorithms is not really a choice because there will only be one version of libtpms in distros and it basically should enable all algorithms (and commands) and make algorithms and commands inaccessible if a certain profile is chosen. So it for example implements TDES but TDES is inaccessible & invisible (TPM2_GetCapability) because the profile doesn't support it.

Understood.
I'm testing configure options. If no option for `--with-only-algs` is supplied, all algorithms are available (or compiled) as normal:
```
./autogen.sh --with-openssl --with-tpm2 --without-tpm1 --with-only-algs=fips140-3 --prefix=/usr && make -j`nproc`
```
Other values would be `fips140-2` or `cnsa1` or `cnsa2`. It can kind of convolute algorithm inclusion in  `TpmProfile.h`, though.

stefanberger:
> Understood. I'm testing configure options. If no option for `--with-only-algs` is supplied, all algorithms are available (or compiled) as normal:
> 
> ```
> ./autogen.sh --with-openssl --with-tpm2 --without-tpm1 --with-only-algs=fips140-3 --prefix=/usr && make -j`nproc`
> ```
> 
> Other values would be `fips140-2` or `cnsa1` or `cnsa2`. It can kind of convolute algorithm inclusion in `TpmProfile.h`, though.

I understand but IMO there will only be one version of libtpms that has as many algorithms enabled as there are possible and the disablement happens via profiles ...

stefanberger:
Cab you prefix your commit messages with 'tpm2:'?

stefanberger:
... and squash the patches into one...

ben-dav-lytle:
> Can you prefix your commit messages with 'tpm2:'?
... and squash the patches into one...

Done!

stefanberger:
Thanks.

PR #369, opened 2023-04-19
tpm2: configure vendor/manufacturer values

ben-dav-lytle:
Provide a way to configure Vendor/Manufacturer values.
This also changes default vendor/manufacturer strings to be more generic (remove IBM specifics; makes IBM less liable? 😄)
Cleaned up comments that were a little confusing in VendorString.h, most likely from the Microsoft Reference code port.

I know a major concern is swtpm state and downgrading versions, and these values (at least Manufacturer, Vendor Strings, and Firmware Version) show up in the properties-fixed get capability command. What I'm not sure of is how tightly this is coupled with 
 the swtpm state and how it would affect earlier libtpms versions.

### DIFF
--- a/src/tpm2/VendorString.h
+++ b/src/tpm2/VendorString.h
@@ -62,47 +62,67 @@
 #ifndef VENDORSTRING_H
 #define VENDORSTRING_H
 
-/* Define up to 4-byte values for MANUFACTURER.  This value defines the response for
-   TPM_PT_MANUFACTURER in TPM2_GetCapability(). The following line should be un-commented and a
-   vendor specific string should be provided here. */
-#define    MANUFACTURER    "IBM"
-
-/*     The following #if macro may be deleted after a proper MANUFACTURER is provided. */
-#ifndef MANUFACTURER
-#error MANUFACTURER is not provided.				 \
-    Please modify VendorString.h to provide a specific	  	 \
-    manufacturer name.
+#ifndef _STRINGIFY
+#define _STRINGIFY(x) #x
+#endif
+#ifndef STRINGIFY
+#define STRINGIFY(x) _STRINGIFY(x)
 #endif
 
-/*     Define up to 4, 4-byte, vendor-specific values. The values must each be 4 bytes long and the
-       last value used may contain trailing zeros. These values define the response for
-       TPM_PT_VENDOR_STRING_(1-4) in TPM2_GetCapability(). The following line should be un-commented
-       and a vendor specific string.  The vendor strings 2-4 may also be defined as appropriate. */
+/*      To customize the MANUFACTURER macro, use the configure option "--with-manufacturer=XXXX"
+        Define up to a 4-byte string for MANUFACTURER.  This string defines the response for
+        TPM_PT_MANUFACTURER in TPM2_GetCapability().
+        Must be in the TPM Vendor ID Registry: https://trustedcomputinggroup.org/resource/vendor-id-registry/ */
+#ifndef CONFIG_MANUFACTURER
+#define MANUFACTURER    "SIM0"
+#else
+#define MANUFACTURER STRINGIFY(CONFIG_MANUFACTURER)
+#endif
+_Static_assert(sizeof(MANUFACTURER) - 1U <= 4U, "MANUFACTURER string can be up to 4-bytes");
 
-#define       VENDOR_STRING_1       "SW  "
-#define       VENDOR_STRING_2       " TPM"
-//#define       VENDOR_STRING_3
-//#define       VENDOR_STRING_4
+/*      To customize the MANUFACTURER_ID macro, use the configure option "--with-manufacturer-id=XXXX"
+        Define up to a 4-byte hex value for MANUFACTURER_ID.  This value defines the response for
+        manufacturer in TPMAttributes for TPM2_GetInfo().
+        Must be in the TPM Vendor ID Registry: https://trustedcomputinggroup.org/resource/vendor-id-registry/ */
+#ifndef MANUFACTURER_ID
+#define MANUFACTURER_ID    0x53494d30
+#endif
+_Static_assert((sizeof(MANUFACTURER_ID) <= 4U), "MANUFACTURER_ID can be up to 4-bytes");
 
-/*     The following #if macro may be deleted after a proper VENDOR_STRING_1 is provided. */
-#ifndef VENDOR_STRING_1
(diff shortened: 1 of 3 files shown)


stefanberger:
> Provide a way to configure Vendor/Manufacturer values.
This also changes default vendor/manufacturer strings to be more generic (remove IBM specifics; makes IBM less liable? 😄)

I am not sure whether this flexibility will be very helpful, though:

By now any the change to any of the fields can cause backwards compatibility issues, such as for example with pkcs11 drivers where the name of the device goes into the pkcs11 URI and an upgrade of libtpms could then be causing the device URI not to work anymore (since the device is identified differently) and crypto to fail. If we want to make it flexible then the default values MUST all be exactly as they are now. Also, migration between differently configured libtpms could cause these types of issues.

ben-dav-lytle:
> causing the device URI not to work anymore (since the device is identified differently) and crypto to fail.

That's what I was afraid of.

> If we want to make it flexible then the default values MUST all be exactly as they are now. Also, migration between differently configured libtpms could cause these types of issues.

If we make the default values the exact same as they are now, I can see one obvious issue with differing libtpms versions - 
[TPM2_GetInfo](https://github.com/stefanberger/libtpms/blob/2ef7f12aedde8b165bf98a6fb59d76e5aafcfccc/src/tpm_tpm2_interface.c#L366) would return all the hardcoded values at that libtpms version (if not at a version where these were configurable). So this output could differ from the TPM2 fixed properties capability.

ben-dav-lytle:
> That would be an unnecessary annoyance for any user that doesn't know why things do not work anymore

Completely agree.

> Who would be configuring libtpms with differing values than what they have been for years now?

If a different company wants to use this project to create a TPM, and brand it as something else (something like Google Cloud: https://cloud.google.com/blog/products/identity-security/virtual-trusted-platform-module-for-shielded-vms-security-in-plaintext, where they use https://sourceforge.net/projects/ibmswtpm2/).

stefanberger:
> If a different company wants to use this project to create a TPM, and brand it as something else (something like Google Cloud: https://cloud.google.com/blog/products/identity-security/virtual-trusted-platform-module-for-shielded-vms-security-in-plaintext, where they use https://sourceforge.net/projects/ibmswtpm2/).

Is this an expressed intention from Google Cloud (or anyone else) or a hypothesis?

src/tpm2/VendorString.h
```diff
@@ -62,47 +62,67 @@
...
 
-/* the more significant 32-bits of a vendor-specific value indicating the version of the firmware
-   The following line should be un-commented and a vendor specific firmware V1 should be provided
-   here. The FIRMWARE_V2 may also be defined as appropriate. */
-#define   FIRMWARE_V1         (0x20191023)
+/*      To customize the VENDOR_STRING_[1-4] macros, use the configure options "--with-vendor-string-[1-4]=XXXX"
+        Define up to 4, 4-byte, vendor-specific strings. The strings must each be 4 bytes long.
+        These values define the response for TPM_PT_VENDOR_STRING_(1-4) in TPM2_GetCapability(). */
+#ifndef CONFIG_VENDOR_STRING_1
+#define VENDOR_STRING_1       "fTPM"
```
stefanberger, line 96: The default here would have to be "SW  "

src/tpm2/VendorString.h
```diff
@@ -62,47 +62,67 @@
...
+#else
+#define VENDOR_STRING_1 STRINGIFY(CONFIG_VENDOR_STRING_1)
+#endif
+_Static_assert(sizeof(VENDOR_STRING_1) - 1U == 4U, "VENDOR_STRING_1 must be 4-bytes");
 
-// the less significant 32-bits of a vendor-specific value indicating the version of the firmware
-#define   FIRMWARE_V2         (0x00163636)
+#ifdef CONFIG_VENDOR_STRING_2
```
stefanberger, line 102: This would have to have a default of " TPM".

src/tpm2/VendorString.h
```diff
@@ -62,47 +62,67 @@
...
-/*     The following #if macro may be deleted after a proper MANUFACTURER is provided. */
-#ifndef MANUFACTURER
-#error MANUFACTURER is not provided.				 \
-    Please modify VendorString.h to provide a specific	  	 \
-    manufacturer name.
+#ifndef _STRINGIFY
+#define _STRINGIFY(x) #x
+#endif
+#ifndef STRINGIFY
+#define STRINGIFY(x) _STRINGIFY(x)
```
stefanberger, line 69: These could be useful elsewhere. Move to a Utils.h?

src/tpm2/VendorString.h
```diff
@@ -62,47 +62,67 @@
...
-// The following #if macro may be deleted after a proper FIRMWARE_V1 is provided.
+/*      To customize the FIRMWARE_V[1-2] macros, use the configure options "--with-firmware-v[1-2]=XXXX"
+        Define the more significant 32-bits of a vendor-specific value indicating the
+        version of the firmware in FIRMWARE_V1. The less significant 32-bits of a vendor-specific
+        value indicating the version of the firmware can use the FIRMWARE_V2 macro. */
 #ifndef FIRMWARE_V1
-#error  FIRMWARE_V1 is not provided.					\
-    Please modify VendorString.h to provide a vendor specific firmware \
-    version
+#define FIRMWARE_V1     0x00000000
```
stefanberger, line 120: Also this here should go back to its previous default value. None of the existing users, like distros, should have to modify their build process and `./configure` line to get back the old default values. So **all** values have to default to what they were before.

ben-dav-lytle:
> > If a different company wants to use this project to create a TPM, and brand it as something else (something like Google Cloud: https://cloud.google.com/blog/products/identity-security/virtual-trusted-platform-module-for-shielded-vms-security-in-plaintext, where they use https://sourceforge.net/projects/ibmswtpm2/).
> 
> Is this an expressed intention from Google Cloud (or anyone else) or a hypothesis?

I believe what Google Cloud has detailed in the article is their current vTPM solution. Other than that, it is a hypothesis. `swtpm` has broad reach, so I wouldn't be surprised if other companies are basing their solutions off it. That's the intention of this patch; an "open-sourced" way to configure the swtpm, if you will.

src/tpm2/VendorString.h
ben-dav-lytle: Found it already defined in `tpm_library_intern.h`

PR #371, opened 2023-06-04
tpm2: Add test vectors for camellia-192

fergus-dall:


stefanberger:
Thanks. This seems to correspond to output of `./src/tpm2/gensymtestsdata.sh`.

Enabling this in libtpms would be more difficult primarily due to compatibility concerns between versions related to migration of the TPM's state. We have Camellia 128 and 256 already enabled and 192 in between won't add much IMO.

fergus-dall:
The compatibility issue is just that after downgrading you wouldn't be able to use objects created with the newer algorithm, right? In that case it should be fine to leave it disabled-by-default, like AES-192, at least until the next version that enables new algorithms by default is released.

Regardless, for my purposes it's fine if it's an optional build config, like AES-192.

PR #377, opened 2023-08-14
Object: clear out sensitive area if on load if not provided

rshearman:
When an object is later marshalled in TPM2_ContextSave, the publicOnly attribute isn't taken into account and therefore potentially stale sensitive information can be marshalled, which is a problem if the buffer sizes it contains have values that are too large - this triggers assertion failures.

Avoid this by clearing out the sensitive area upon ObjectLoad if not provided, making the behaviour consistent with when a fresh, unused, object entry is used.


src/tpm2/Object.c
```diff
@@ -460,7 +460,10 @@ ObjectLoad(
     object->publicArea = *publicArea;
     // If there is a sensitive area, load it
     if(sensitive == NULL)
-	object->attributes.publicOnly = SET;
+	{
+	    object->attributes.publicOnly = SET;
+	    MemorySet(&object->sensitive, 0, sizeof(object->sensitive));
```
stefanberger, line 465: This line is necessary due to modifications by libtpms that affect how the state of OBJECTs is marshalled. Can you add a comment here `// needed for libtpms`.

Thank you!

src/tpm2/Object.c
```diff
@@ -460,7 +460,10 @@ ObjectLoad(
     object->publicArea = *publicArea;
     // If there is a sensitive area, load it
     if(sensitive == NULL)
-	object->attributes.publicOnly = SET;
+	{		// libtpms changed begin
+	    object->attributes.publicOnly = SET;
+	    MemorySet(&object->sensitive, 0, sizeof(object->sensitive)); // needed for libtpms.
```
stefanberger, line 465: Actually, are you sure that this is necessary? Do you have a test case? I am asking because `ObjectLoad()` is called with an object pointer from two locations:

```
-> src/tpm2/ObjectCommands.c:    result = ObjectLoad(newObject, parentObject,
-> src/tpm2/ObjectCommands.c:    result = ObjectLoad(object, NULL,
not this one: src/tpm2/DuplicationCommands.c:     result = ObjectLoad(NULL, NULL, &in->objectPublic.publicArea,
```

Both get the object they pass into this function like this:
```
    newObject = FindEmptyObjectSlot(&out->objectHandle);

    object = FindEmptyObjectSlot(&out->objectHandle);
```

FindEmptyObjectSlot has been instrumented previously to clear the object:

```
OBJECT *
FindEmptyObjectSlot(
                    TPMI_DH_OBJECT  *handle         // OUT: (optional)
                    )
{
    UINT32               i;
    OBJECT              *object;
    for(i = 0; i < MAX_LOADED_OBJECTS; i++)
        {
            object = &s_objects[i];
            if(object->attributes.occupied == CLEAR)
                {
                    if(handle)
                        *handle = i + TRANSIENT_FIRST;
                    // Initialize the object attributes
                    // MemorySet(&object->attributes, 0, sizeof(OBJECT_ATTRIBUTES));
                    MemorySet(object, 0, sizeof(*object)); // libtpms added: Initialize the whole object
                    return object;
                }
        }
    return NULL;
}
```

stefanberger:
Unless there's something wrong with the existing clearing of the OBJECT, I will revert this patch.

src/tpm2/Object.c
rshearman: Apologies, you are absolutely right. I was using a (much) older version that didn't have 17255da54cf8354d02369f1323dc50cfb87e2bf4 and that addresses the problem I was seeing. This can be reverted.

PR #395, opened 2024-01-05
Sync the RSA prime number generation algorithm with rev180

stefanberger:
This PR syncs the RSA prime number generation algorithm with rev180. Since it now uses a different random number generation function the usage of the new algorithm has to be made dependent on the SEED_COMPAT_LEVEL.
Also, update the sieveMarks array with the missing member and roll back the previously made change to avoid mark=5. Since 3072 bit RSA keys may occasionally be different if the full sieveMarks array is now use, we also have to make the usage of the full (or restricted) sieveMarks array dependent on the SEED_COMPAT_LEVEL.

### DIFF
--- a/src/tpm2/TpmMath_Util.c
+++ b/src/tpm2/TpmMath_Util.c
@@ -99,6 +99,77 @@ LIB_EXPORT BOOL TpmMath_IntTo2B(
     return FALSE;
 }
 
+//*** TpmMath_GetRandomBits()
+// This function gets random bits for use in various places.
+//
+// One consequence of the generation scheme is that, if the number of bits requested
+// is not a multiple of 8, then the high-order bits are set to zero. This would come
+// into play when generating a 521-bit ECC key. A 66-byte (528-bit) value is
+// generated and the high order 7 bits are masked off (CLEAR).
+// In this situation, the highest order byte is the first byte (big-endian/TPM2B format)
+//  Return Type: BOOL
+//      TRUE(1)         success
+//      FALSE(0)        failure
+LIB_EXPORT BOOL TpmMath_GetRandomBits(BYTE* pBuffer, size_t bits, RAND_STATE* rand)
+{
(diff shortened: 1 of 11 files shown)


src/tpm2/TpmMath_Util.c
```diff
@@ -99,6 +99,78 @@ LIB_EXPORT BOOL TpmMath_IntTo2B(
...
+	    // if the number of bits % 8 != 0, mask the high order (first) byte to the relevant number of bits
+	    // bits % 8     desired mask   right-shift of 0xFF
+	    //     0           0xFF             0 = (8 - 0) % 8
+	    //     1           0x01             7 = (8 - 1) % 8
+	    //     2           0x03             6 = (8 - 2) % 8
+	    //     ... etc ...
+	    //     7           0x7F             1 = (8 - 7) % 8
+	    int  excessBits = bits % 8;
+	    int  shift      = (8 - excessBits) % 8;
+	    BYTE mask       = ~(0xFF >> shift);
```
stefanberger, line 167: The `~` here looks wrong.

stefanberger:
Assessment of patches:

- [] tpm2: Adjust primeLimit depending on SEED_COMPAT_LEVEL
- [] rev180: Add missing entry to sieveMarks array
- [] rev180: Add new code to generate an RSA prime => introduce SEED_COMPAT_LEVEL_RSA_PRIME_ADJUST_REV165
- [] tpm2: Introduce SEED_COMPAT_LEVEL_RSA_PRIME_ADJUST_PRE165 and use it => a renaming; looks ok
- [] tpm2: Move code out of RsaAdjustPrimeCandidate => looks ok
- [] tpm2: Append suffix _PreRev165 to RsaAdjustPrimeCandidate => Just a renaming of function; harmless
- [] rev180: Add TpmMath_GetRandomBits to be used for RSA prime generation => bugfix applied; needs testing; harmless at this point

Names of seed levels need fixing.

PR #407, opened 2024-03-13
tpm12: add missing openssl includes

orbea:
This fixes the build with LibreSSL 3.9.0 where many implicit declarations for `BN_`, `EVP_` and `RSA_` functions occur which were implicitly included before.

Also, `make check` passes for me with LibreSSL 3.9.0.

### DIFF
--- a/src/tpm12/tpm_crypto.c
+++ b/src/tpm12/tpm_crypto.c
@@ -48,6 +48,8 @@
 #include <openssl/rand.h>
 #include <openssl/sha.h>
 #include <openssl/engine.h>
+#include <openssl/evp.h>
+#include <openssl/rsa.h>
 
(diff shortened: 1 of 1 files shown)


src/tpm12/tpm_crypto.c
```diff
@@ -48,6 +48,8 @@
 #include <openssl/rand.h>
 #include <openssl/sha.h>
 #include <openssl/engine.h>
+#include <openssl/evp.h>
+#include <openssl/rsa.h>
```
stefanberger, line 52: Can you prefix your patch description with `tpm12:`, please. Otherwise LGTM.

orbea: Done.

PR #424, opened 2024-07-16
Implement KDFe using OpenSSL's SSKDF

stefanberger:


stefanberger:
@elmarco Any suggestions on this PR here and the test case verifying that the two implementations provide the same result? I am keeping the test case around though we could probably do without it (entirely).

PR #426, opened 2024-07-17
DO-NOT-MERGE: Implement TPM 2's KDFa using OpenSSL KBKDF

stefanberger:
This PR currently builds on top of the KDFe patches/PR.

Implement TPM 2's KDFa using OpenSSL KBKDF. However, due to the flexibility of the reference implementation (allow to 'resume' the KDFe at a later point) and restrictions by OpenSSL's KBKDF it will not be possible to use the OpenSSL-based implementation -> DO NOT MERGE! See commit messages on ''tests: Add test case for KDFa replacement by OpenSSL 'KBKDF'".

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,7 @@ Makefile
 /tests/base64decode
 /tests/fuzz
 /tests/freebl_sha1flattensize
+/tests/kdfe_openssl
 /tests/nvram_offsets
 /tests/object_size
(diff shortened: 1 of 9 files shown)


stefanberger:
@elmarco FYI: Unless we significantly modify the TPM 2 code and how it uses KDFa I don't see how we can replace it with an OpenSSL implementation, if this was necessary.

PR #428, opened 2024-07-18
Add support for FIPS restrictions on more crypto algorithms

stefanberger:
This PR adds support for a FIPS restrictions on algorithms. To do this it introduces an `Attribute` field in the TPM profile where a comma separated list of attributes can be placed. 

One such attribute is `fips-host` that 'restricts' some crypto algorithms of the TPM 2 in the same way as they are restricted by OpenSSL (reference point is RHEL 9.4+): no raw RSA encryption and decryption, forbids sha1 signature generation, signing with RSA < 2048, and key generation for RSA < 2048.

Another attribute is `fips-140-3` that imposes yet more 'restrictions' on crypto algorithms and activates pairwise consistency tests etc. This part is not complete.


stefanberger:
A few questions:
- Is this series useful? 
- Is this series complete to make any FIPS-140-3 claims?

PR #462, opened 2024-09-30
tpm2: Enable DRBG continous test: drbg-continous-test

stefanberger:
drbg-continous-test enables an existing code block that was previously only enabled when FIPS_COMPLIANT #define was set. This code block ensures that previous 4 consecutive random numbers do not appear again at the beginning of a 16-byte block.

Extend an existing test case with this new attribute.


stefanberger:
Yet to run a TPM 2 with the DRBG continous testing (FIPS_COMPLIANT) flag enabled.

stefanberger:
The TPM 2 code for continuous testing is functioning correctly.

https://csrc.nist.gov/files/pubs/fips/140-2/upd2/final/docs/fips1402.pdf

```
Continuous random number generator test. If a cryptographic module employs Approved or non-
Approved RNGs in an Approved mode of operation, the module shall perform the following continuous
random number generator test on each RNG that tests for failure to a constant value.
1. If each call to a RNG produces blocks of n bits (where n > 15), the first n-bit block generated
after power-up, initialization, or reset shall not be used, but shall be saved for comparison with
the next n-bit block to be generated. Each subsequent generation of an n-bit block shall be
compared with the previously generated block. The test shall fail if any two compared n-bit
blocks are equal.
2. If each call to a RNG produces fewer than 16 bits, the first n bits generated after power-up,
initialization, or reset (for some n > 15) shall not be used, but shall be saved for comparison with
the next n generated bits. Each subsequent generation of n bits shall be compared with the
previously generated n bits. The test fails if any two compared n-bit sequences are equal.
```

PR #469, opened 2024-10-15
WIP: tpm2: Add RuntimeAttribute fips-140-3 to restrict TPM 2 (StateFo…

stefanberger:
…rmatLevel 8)

Add RuntimeAttribute 'fips-140-3' that restricts a TPM 2 following requirements of the FIPS-140-3 standard:

- Prevents raw RSA encryption and decryption
- Prevents SHA1 signature generation + verification
- Prevents derivation of ECC keys from derivation parent and for creation of ephemeral keys
- Performs a pairwise consistency test on RSA keys and ECC signing keys
- Enables continous DRBG test


stefanberger:
We can only enable the `fips-140-3` attribute when we know that support for this is 'complete'. Problem is I do not know whether FIPS-140-3 support 'is complete' as it is implemented.

PR #470, opened 2024-10-17
Add support for enabling the SVN-limited hierarchy

stefanberger:
Add support for the SVN-limited hierarchy and allow enabling it with a profile attribute svn-limited-hierarchy.

The SVN limited hierarchy is based on a 16 bytes SVN base secret that is unique per vTPM instance and is created as a random number on first manufacturing.


src/tpm_tpm2_interface.c
```diff
@@ -871,6 +871,13 @@ static TPM_BOOL TPM2_WasManufactured(void)
     return g_wasManufactured;
 }
 
+static TPM_RESULT TPM2_RecreateSvnBaseSecret(void)
+{
+    if (_plat__SvnBaseSecretRecreate() == 0)
```
stefanberger, line 876: Should probably trigger the storing of the persistent state.

PR #486, opened 2025-04-24
Enable support for 4096-bit RSA keys by default

ddstreet:
Currently, the defaults do not allow 4096-bit RSA keys, however many end users are starting to use them. The lack of support in libtpms - configured at compile time, not runtime - means that most Linux distros (which don't currently alter the defaults) do not allow the use of 4096-bit RSA keys inside virtual machines.

Not only does this prevent creating 4096-bit RSA keys in the tpm (e.g. tpm2_createprimary), but it also prevents loading external 4096-bit RSA keys into the VM's tpm, for example when attempting to use the tpm to verify an externally-generated data signature that was created using a 4096-bit RSA key (e.g. as part of a PolicyAuthorize and/or VerifySignature).

### DIFF
--- a/src/tpm2/RuntimeProfile.c
+++ b/src/tpm2/RuntimeProfile.c
@@ -78,7 +78,7 @@ static const struct RuntimeProfileDesc {
      * This basically locks the name of the profile to the stateFormatLevel.
      */
     unsigned int stateFormatLevel;
-#define STATE_FORMAT_LEVEL_CURRENT 7
+#define STATE_FORMAT_LEVEL_CURRENT 8
 #define STATE_FORMAT_LEVEL_UNKNOWN 0 /* JSON didn't provide StateFormatLevel; this is only
 					allowed for the 'default' profile or when user
 					passed JSON via SetProfile() */
@@ -105,6 +105,7 @@ static const struct RuntimeProfileDesc {
  *      - drbg-continous-test
  *      - pct
  *      - no-ecc-key-derivation
+ *  8 : Enabled 4096-bit RSA support
  */
     const char *description;
 #define DESCRIPTION_MAX_SIZE        250
@@ -971,8 +972,8 @@ RuntimeProfileGetSeedCompatLevel(void)
     case 1: /* profile runs on v0.9 */
 	return SEED_COMPAT_LEVEL_RSA_PRIME_ADJUST_FIX;
 
-    case 2 ... 7: /* profile runs on v0.10 */ {
-	MUST_BE(STATE_FORMAT_LEVEL_CURRENT == 7); // force update when this changes
+    case 2 ... 8: /* profile runs on v0.10 */ {
+	MUST_BE(STATE_FORMAT_LEVEL_CURRENT == 8); // force update when this changes
 	return SEED_COMPAT_LEVEL_LAST;
     }
 
--- a/src/tpm2/RuntimeProfile.c
+++ b/src/tpm2/RuntimeProfile.c
@@ -111,9 +111,27 @@ static const struct RuntimeProfileDesc {
 #define DESCRIPTION_MAX_SIZE        250
     bool allowModifications; /* user is allowed to modify algorithms profile */
 } RuntimeProfileDescs[] = {
-#define PROFILE_DEFAULT_IDX	0
-#define PROFILE_NULL_IDX	1
-    [PROFILE_DEFAULT_IDX] = {
+#define PROFILE_DEFAULT_V2_IDX	0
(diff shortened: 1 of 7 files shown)


stefanberger:
This will take more than just these changes. At least the following will also be necessary:
- create a default-v2 profile from the current default-v1 profile
- bump up the StateFormatLevel and enable RSA_4096 in RuntimeAlgorithms.c
- ...

I was going to wait for PQC support to become available for a new version of libtpms. In the meantime I would try to use RSA-3072 or NIST-P256/384/521 for signatures.

ddstreet:
> This will take more than just these changes. At least the following will also be necessary:
> 
> * create a default-v2 profile from the current default-v1 profile
> * bump up the StateFormatLevel and enable RSA_4096 in RuntimeAlgorithms.c
> * ...
> 
> I was going to wait for PQC support to become available for a new version of libtpms. In the meantime I would try to use RSA-3072 or NIST-P256/384/521 for signatures.

I updated the PR as suggested, as well as I could, but I may have missed something, please feel free to exit this PR as needed, or if you would prefer to just fix it with your own commit please feel free.

Also I opened #491 in case you want to fix this with your own commit(s), and can track it through that issue.

ddstreet:
Also, it seems like this isn't something that would be safely backported in existing stable distros, due to the use of strict versioning of the profiles? Meaning, it wouldn't really make sense to enable 4096-bit RSA support in existing distro 0.10 (or earlier) packages? Basically, it will require a full update to 0.11 (when that's released)?

stefanberger:
> Also, it seems like this isn't something that would be safely backported in existing stable distros, due to the use of strict versioning of the profiles? Meaning, it wouldn't really make sense to enable 4096-bit RSA support in existing distro 0.10 (or earlier) packages? Basically, it will require a full update to 0.11 (when that's released)?

I would rather NOT want to backport anything related to these profiles and especially the StateFormatLevel but have this only be part of v0.11 and only enabled in default-v2 or custom profiles starting with StateFormatLevel 8. libtpms v0.10 and default-v1 profile are not expected to support this. It's tricky the way it is with downgradability while new features are being added. I would also like to avoid making too many profiles that don't mean much to users, meaning wait until it's worth releasing a new version that has something significantly new in the default-v2 profile.

I saw you commit. I would like to have the default-v2 profile creation in a separate patch. If you want to iterate over this with me for a while it's fine, otherwise I will look at it myself at some point.

src/tpm2/RuntimeProfile.c
```diff
@@ -111,9 +111,27 @@ static const struct RuntimeProfileDesc {
...
+	.commandsProfile   = "0x11f-0x122,0x124-0x12e,0x130-0x140,0x142-0x159,0x15b-0x15e,"
+			     "0x160-0x165,0x167-0x174,0x176-0x178,0x17a-0x193,0x197,"
+			     "0x199-0x19c",
+	.algorithmsProfile = "rsa,rsa-min-size=1024,tdes,tdes-min-size=128,sha1,hmac,"
+			     "aes,aes-min-size=128,mgf1,keyedhash,xor,sha256,sha384,sha512,"
+			     "null,rsassa,rsaes,rsapss,oaep,ecdsa,ecdh,ecdaa,sm2,ecschnorr,ecmqv,"
+			     "kdf1-sp800-56a,kdf2,kdf1-sp800-108,ecc,ecc-min-size=192,ecc-nist,"
+			     "ecc-bn,ecc-sm2-p256,symcipher,camellia,camellia-min-size=128,cmac,"
+			     "ctr,ofb,cbc,cfb,ecb",
+	.stateFormatLevel  = 8,
```
stefanberger, line 129: At this point it should be '7'.

ddstreet: ack, updated

ddstreet:
just checking in, does this look ok?

stefanberger:
> just checking in, does this look ok?

Haven't gotten around to look at it, yet. Soon...

stefanberger:
Can you put `tpm2:` to each one of the commit titles and write something about the default-v2 profile that it's basically a copy of the default-v1 profile at this point? If it passes the tests, I will merge.

ddstreet:
> Can you put `tpm2:` to each one of the commit titles and write something about the default-v2 profile that it's basically a copy of the default-v1 profile at this point? If it passes the tests, I will merge.

ack, updated thanks.

stefanberger:
Tests with swtpm should pass once this has been merged: https://github.com/stefanberger/swtpm/pull/1019

PR #527, opened 2025-08-30
add pkg-config to ubunto prerequisites

ernst-schwaiger:
On Ubuntu 24.04 pkg-config is not installed per default, but it is needed to run `./autogen.sh` without errors.


stefanberger:
Please add your Signed-off-by to your commit.

PR #528, opened 2025-08-31
add pkg-config to ubuntu prerequisites

ernst-schwaiger:
Signed-off-by: ernst.schwaiger@chello.at


stefanberger:
I don't know why it still complains about the DCO. Thanks.

PR #529, opened 2025-09-22
test: do not run localcheck test on powerpc64

bluca:
There is a spurious failure on powerpc64:

$ gcc -nostdlib -Lsrc/.libs -ltpms
/usr/bin/ld: src/.libs/libtpms.so: ABI version 1 is not compatible with ABI version 0 output
/usr/bin/ld: failed to merge target specific data of file src/.libs/libtpms.so
/usr/bin/ld: warning: cannot find entry symbol _start; defaulting to 00000000000000d8
collect2: error: ld returned 1 exit status

This is likely a linker bug, which has been reported, but
even if it is fixed, it will take a long time before it is available:

https://sourceware.org/bugzilla/show_bug.cgi?id=17742#c7

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -5,7 +5,7 @@
 #
 
 AM_CFLAGS = @AM_CFLAGS@
-AM_LDFLAGS = @AM_LDFLAGS@ $(HARDENING_LDFLAGS) $(SANITIZERS) $(FUZZER) 
+AM_LDFLAGS = @AM_LDFLAGS@ $(HARDENING_LDFLAGS) $(SANITIZERS) $(FUZZER)
 noinst_LTLIBRARIES =
 noinst_HEADERS =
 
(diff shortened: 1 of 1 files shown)


bluca:
> Can you please also describe in which distro and version you are encountering this error? I am running tests also on ppc64 but not across a whole lot of distros. I haven't encountered this issue so far on ppc64. Also, if it's a linker issue I would expect that it will be patched rather quickly. Also, the distro encountering this issue should carry this patch until the linker is fixed.

Latest version building in Debian testing/unstable, and it has been reported on Ubuntu too. I assumed others would run in the same issue and hence I sent the PR.

> Also, in your error report to sourceware.org you picked up on a bug originally from 2015, which I would expect that it has been fixed a long time ago.

Yes I found that by googling for the error, and it's the only hit. Given the message and the affected architecture is the same, it looked too much to be a concidence, so I reported it there. It's not exactly the same, as the original issue was solved, so it's likely a slightly different but related corner case - but I am not a compiler expert, so just speculating.

stefanberger:
If this is still an issue, then we can reopen it. I haven't seen it on ppc64 on Fedora...

PR #584, opened 2026-05-28
fix(RuntimeProfile): Fix false deduplication of tokens that are a suffix of last token.

IWillBurn:
Fix false deduplication of tokens that are a suffix of last token.

When processing a comma-separated list containing `<token>` at an arbitrary location and `other-<token>` at the end of the list, the `RuntimeProfileDedupStrItems` function incorrectly deletes `<token>` as a duplicate.

The root cause is in the boundary check when a `strstr` found match at the end of the string:

```c
if (((dup[-1] == ',' || dup[-1] == 0) && dup[slen] == exp) ||
     (dup[slen] == 0) /* last item in list */)
```

When `<token>` is matched as a substring inside `other-<token>` (the last item), the `(dup[slen] == 0)` branch triggers without verifying that the left boundary of the match is valid — i.e. that it is preceded by a `,` or start of string. This causes `<token>` to be incorrectly deduplicated out of the list.

The fix tightens the boundary check to require a valid left-side delimiter in all cases:

```c
if ((dup[-1] == ',' || dup[-1] == 0) &&
     (dup[slen] == exp || dup[slen] == 0))
```

This ensures a match is only treated as a true duplicate when both the left boundary and the right boundary are valid.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved comma-separated list deduplication: duplicates are now removed only when an item has valid boundaries on both sides, preventing partial-match edge cases and reducing incorrect removals when processing in-place comma-delimited values.

### DIFF
--- a/src/tpm2/RuntimeProfile.c
+++ b/src/tpm2/RuntimeProfile.c
@@ -266,9 +266,9 @@ RuntimeProfileDedupStrItems(char *input)
         while (true) {
             dup = strstr(ncomma + 1, pos);
             if (dup) {
-                /* ensure 'dup' is a prefix of 'pos' with either ',' or '\0' before it */
-                if (((dup[-1] == ',' || dup[-1] == 0) && dup[slen] == exp) ||
-                     (dup[slen] == 0) /* last item in list */) {
+                /* ensure 'dup' is a whole token: valid left boundary AND right boundary */
+                if ((dup[-1] == ',' || dup[-1] == 0) &&
+                    (dup[slen] == exp || dup[slen] == 0)) {
                     memmove(pos, comma + 1, strlen(comma + 1) + 1);
                     /* keep pos as-is */
                     found = true;


stefanberger:
Can you update your patch description and state in which case the problem occurred. This is useful for later on as well.

So I guess in your case you have this here as you said: "...,gost3410,ecc-tc26-gost3410"

The deduplication algo looks at `gost3410` and tries to find a duplicate of this string following it. It finds a strstr-match inside `ecc-tc26-gost3410`. Because `ecc-tc26-gost3410` happens to be the last one in the string we do not succeed on comparing with `(dup[-1] == ',' || dup[-1] == 0) && dup[slen] == exp)` but we go by `(dup[slen] == 0)` [last one in string is followed by `\0`]. Ok, I can see that this is a bug.

Your fix:
```
               if ((dup[-1] == ',' || dup[-1] == 0) &&
                    (dup[slen] == exp || dup[slen] == 0)) {
```

With your fix it wouldn't match the dup[-1] and we don't even need to look at the check after the `&&`. Alright, I agree, this is a good fix. Thanks.

IWillBurn:
Yes, that's right, that's the problem. Is there anything else that needs to be done to get the PR accepted? It seems "coverage/coveralls" is not working, maybe it should be restarted or something else.

stefanberger:
> Is there anything else that needs to be done to get the PR accepted?

 A description of what the patch fixes would be good.

IWillBurn:
Done.

stefanberger:
> Done.

The text should go into the patch description.

stefanberger (approved):

stefanberger:
Thanks!

PR #588, opened 2026-06-10
tpm2: fix GCC 15 stringop-overflow error in MakeIv

baloo:
When compiling with GCC 15 using `CFLAGS=-march=x86-64-v4`, the compiler's aggressively optimized vectorizer triggers a false-positive -Wstringop-overflow error. Because x86-64-v4 enables wide AVX-512 registers, the compiler misinterprets the loop unrolling and warns that a 64-byte vector write is overflowing the destination buffer:
```
  tpm2/TPMCmd/tpm/src/crypt/AlgorithmTests.c:158:17: error: 
  writing 64 bytes into a region of size 15 [-Werror=stringop-overflow=]
    158 |             *iv = i;
```

This fixes the warning by marking the `iv` output pointer parameter as `volatile`. This inhibits the over-aggressive loop vectorization on this specific buffer, silencing the compiler error without changing the underlying logic.

## Summary by CodeRabbit

**No user-visible changes**

* **Refactor**
  * Internal adjustment to how an IV buffer is handled to improve compiler/memory behavior; no functional changes or public interface impacts.

### DIFF
--- a/src/tpm2/TPMCmd/tpm/src/crypt/AlgorithmTests.c
+++ b/src/tpm2/TPMCmd/tpm/src/crypt/AlgorithmTests.c
@@ -138,9 +138,9 @@ TestSMAC(
 
 //*** MakeIv()
 // Internal function to make the appropriate IV depending on the mode.
-static UINT32 MakeIv(TPM_ALG_ID mode,  // IN: symmetric mode
-                     UINT32     size,  // IN: block size of the algorithm
-                     BYTE*      iv     // OUT: IV to fill in
+static UINT32 MakeIv(TPM_ALG_ID     mode,  // IN: symmetric mode
+                     UINT32         size,  // IN: block size of the algorithm
+                     volatile BYTE* iv     // OUT: IV to fill in; 'volatile' for gcc15
 )
 {
     BYTE i;


baloo:
This was reported by https://github.com/NixOS/nixpkgs/issues/528643

src/tpm2/TPMCmd/tpm/src/crypt/AlgorithmTests.c
```diff
@@ -138,9 +138,9 @@ TestSMAC(
 
 //*** MakeIv()
 // Internal function to make the appropriate IV depending on the mode.
-static UINT32 MakeIv(TPM_ALG_ID mode,  // IN: symmetric mode
-                     UINT32     size,  // IN: block size of the algorithm
-                     BYTE*      iv     // OUT: IV to fill in
+static UINT32 MakeIv(TPM_ALG_ID     mode,  // IN: symmetric mode
+                     UINT32         size,  // IN: block size of the algorithm
+                     volatile BYTE* iv     // OUT: IV to fill in
```
stefanberger: May want to add to comment `... OUT: IV to fill in; 'volatile' for gcc15`

baloo:
I will admit I don't understand the test failure, patch doesn't touch a piece swtpm relies upon. Could that be a fluke? (can we retry?)

stefanberger:
> I will admit I don't understand the test failure, patch doesn't touch a piece swtpm relies upon. Could that be a fluke? (can we retry?)

The timeouts that occurred are strange but unrelated to this patch. I will merge it soon. 

Thanks!

PR #611, opened 2026-09-02
crypto: clear generated RSA/ECC private key material before return

### LINKED ISSUE #610: Freshly-generated RSA/ECC private key material never cleared in CryptRsaGenerateKey / CryptEccGenerateKey
## Summary

`CryptRsaGenerateKey()` and `CryptEccGenerateKey()` (`src/tpm2/TPMCmd/tpm/src/crypt/CryptRsa.c`, `CryptEccMain.c`) generate a fresh asymmetric private key into a stack-local buffer and return without clearing it. The freshly-generated private key material — the exact thing these functions exist to protect — is left resident in the stack frame after return.

## RSA: `CryptRsaGenerateKey()`

```c
NEW_PRIVATE_EXPONENT(Z);
```

expands (`include/private/CryptRsa.h`) to:

```c
privateExponent  _Z;
privateExponent* Z = RsaInitializeExponent(&_Z);
```

`_Z` is a stack struct (`entries[5]`, backing `P`/`Q`/`dP`/`dQ`/`qInv`) that holds the two generated RSA primes and their CRT parameters — the actual private key — for the duration of key generation. Every return path (`return retVal;` at the end, and the `ERROR_EXIT`/`Exit:` paths for the validation-failure cases) leaves `_Z` untouched.

## ECC: `CryptEccGenerateKey()`

```c
CRYPT_ECC_NUM(bnD);
```

expands (`MathLibraryInterfaceTypes.h`) to:

```c
CRYPT_INT_BUF(ci_bnD_buf_t, ECC_BITS) bnD_buf;
Crypt_Int* bnD = ExtMath_Initialize_Int((Crypt_Int*)&bnD_buf, ECC_BITS);
```

`bnD_buf` holds the generated private scalar `d`. Same pattern: the `Exit:` path (`CRYPT_CURVE_FREE(E); return retVal;`) never clears it.

## Suggested fix

Clear both buffers right before return, with a call the compiler can't reason its way out of eliding (a plain `memset`/`MemorySet` — `MemorySet` here is a one-line `memset` wrapper, see `src/tpm2/TPMCmd/tpm/src/support/Memory.c` — is a legitimate dead-store-elimination target once the buffer isn't read again). Since this project already links OpenSSL (`USE_OPENSSL_FUNCTIONS_RSA`/`USE_OPENSSL_FUNCTIONS_EC`), `OPENSSL_cleanse()` is available and gives that guarantee without introducing a new dependency:

```c
OPENSSL_cleanse(Z, sizeof(*Z));           /* in CryptRsaGenerateKey, before return */
OPENSSL_cleanse(&bnD_buf, sizeof(bnD_buf)); /* in CryptEccGenerateKey, before return */
```

I'm happy to submit a PR for this — wanted to check first in case there's a preferred zeroing convention/helper for the project (I noticed `MemorySet` is used pervasively elsewhere for hardening zeroization, e.g. `FindEmptyObjectSlot`'s full-object clear in `Object.c`, but a plain `memset` wrapper doesn't carry the same "can't be optimized away" guarantee that key material specifically needs).

## Scope note

This is scoped to the two key-generation entry points, where I could verify concretely (via macro expansion) that the buffer in question is genuinely stack-resident private key material and that no existing code clears it. I did not do a full sweep of every RSA/ECC crypto wrapper in this pass (e.g. sign/decrypt/ECDH intermediate buffers) — happy to look further if useful.

---
> Found and reported with assistance from Claude Code (claude.ai/code).

Johnsonajibi:
## Summary

Fixes #610. `CryptRsaGenerateKey()` and `CryptEccGenerateKey()` generate a fresh asymmetric private key into a stack-local buffer and return without clearing it, leaving the freshly-generated private key material resident in the stack frame after return.

**RSA** (`CryptRsa.c`): `NEW_PRIVATE_EXPONENT(Z)` expands to a stack `privateExponent _Z` struct holding the generated primes P, Q and their CRT parameters (dP, dQ, qInv) — the actual private key.

**ECC** (`CryptEccMain.c`): `CRYPT_ECC_NUM(bnD)` expands to a stack `bnD_buf` holding the generated private scalar `d`.

## Fix

- `CryptRsaGenerateKey()`: add `OPENSSL_cleanse(Z, sizeof(*Z))` at the single return point, which every path funnels through (including the OpenSSL key-generation path, where `Z` is never populated, and the various `ERROR_EXIT`/`Exit:` paths).
- `CryptEccGenerateKey()`: add `OPENSSL_cleanse(&bnD_buf, sizeof(bnD_buf))` at the `Exit:` label. One early `return TPM_RC_FAILURE` inside the PWCT (pairwise consistency test) block bypassed `Exit:` entirely; changed it to set `retVal` and `goto Exit` instead so this path also clears `bnD_buf` rather than skipping cleanup.

`OPENSSL_cleanse()` rather than `MemorySet()`/`memset()`, because the latter (a one-line wrapper, see `src/tpm2/TPMCmd/tpm/src/support/Memory.c`) is a legitimate dead-store-elimination target once the buffer isn't read again. Both files already link OpenSSL (`USE_OPENSSL_FUNCTIONS_RSA` / `USE_OPENSSL_FUNCTIONS_EC`), so this adds no new dependency — `CryptEccMain.c` needed one new `#include <openssl/crypto.h>`, matching what `CryptRsa.c` already has.

## Verification

- `make` (the project's `-Wall -Werror -Wshadow -Wreturn-type -Wsign-compare` build) compiles both files clean.
- `tests/tpm2_createprimary` — the project's own test, which drives a real `TPM2_CreatePrimary` (RSA) command through `TPMLIB_Process()` end-to-end, exercising `CryptRsaGenerateKey()` — passes:
  ```
  OK
  ```
- I traced the ECC side's correctness via macro expansion and scope analysis (`bnD_buf` is declared at the top level of the function body, so it's in scope at every `Exit:`/`goto Exit` site including the one I changed) rather than a live ECC `CreatePrimary` run, since the project's bundled tests don't include an ECC key-generation case I could reuse directly. Happy to add one, or to adjust the fix, if that's wanted.

---
> Found and fixed with assistance from Claude Code (claude.ai/code).

## Summary by CodeRabbit

* **Security**
  * Improved protection for sensitive cryptographic key material by clearing temporary ECC and RSA values after key generation.
  * ECC key generation now follows consistent cleanup handling when random-number generation fails.
  * Strengthened cleanup across successful and unsuccessful key-generation paths to reduce residual sensitive data in memory.
  * Private key components and intermediate calculation data are cleared more reliably before cryptographic operations finish.

### DIFF
--- a/src/tpm2/TPMCmd/tpm/src/crypt/CryptEccMain.c
+++ b/src/tpm2/TPMCmd/tpm/src/crypt/CryptEccMain.c
@@ -7,6 +7,8 @@
 #include "TpmEcc_Signature_ECDSA_fp.h"  // required for pairwise test in key generation
 #include "Helpers_fp.h"                // libtpms added
 
+#include <openssl/crypto.h>  // libtpms added
+
 #if ALG_ECC
 //** Functions
 
@@ -478,6 +480,12 @@ BOOL TpmEcc_GenPrivateScalar(
     OK = OK && ExtMath_Mod(bnExtraBits, nMinus1);
     OK = OK && ExtMath_AddWord(dOut, bnExtraBits, 1);
 
+    // libtpms added begin: bnExtraBits_buf holds the raw random value that
+    // dOut (the private scalar) is directly derived from; clear it here
+    // rather than leaving it in this stack frame.
+    OPENSSL_cleanse(&bnExtraBits_buf, sizeof(bnExtraBits_buf));
+    // libtpms added end
+
     return OK && !_plat__InFailureMode();
 }
 #else                                  // libtpms added begin
@@ -512,6 +520,11 @@ BOOL TpmEcc_GenPrivateScalar(
     OK = OK && ExtMath_SubtractWord(nMinus1, order, 1);
     OK = OK && ExtMath_Mod(bnExtraBits, nMinus1);
     OK = OK && ExtMath_AddWord(dOut, bnExtraBits, 1);
+
+    // bnExtraBits_buf holds the raw random value that dOut (the private
+    // scalar) is directly derived from; clear it here rather than
+    // leaving it in this stack frame.
+    OPENSSL_cleanse(&bnExtraBits_buf, sizeof(bnExtraBits_buf));
 
     return OK && !_plat__InFailureMode();
 }
@@ -573,6 +586,10 @@ BOOL TpmEcc_GenerateKeyPair(Crypt_Int*            bnD,  // OUT: private scalar
     } else {
         OK = OK && ExtEcc_PointMultiply(ecQ, NULL, bnD, E);
     }
+    // bnD1_buf is bnD + order (order is public), so it trivially reveals
+    // the private scalar bnD; clear it here rather than leaving it in
+    // this stack frame.
+    OPENSSL_cleanse(&bnD1_buf, sizeof(bnD1_buf));
     return OK;
 }
 
@@ -754,7 +771,8 @@ LIB_EXPORT TPM_RC CryptEccGenerateKey(
         // Get a random value to sign using the built in DRBG state
         DRBG_Generate(NULL, digest.t.buffer, digest.t.size);
         if(_plat__InFailureMode())
-            return TPM_RC_FAILURE;
+            ERROR_EXIT(TPM_RC_FAILURE); // libtpms changed: fall through to
+                                        // Exit so bnD_buf is cleared below
         TpmEcc_SignEcdsa(bnT, bnS, E, bnD, &digest, NULL);
         // and make sure that we can validate the signature
         OK = TpmEcc_ValidateSignatureEcdsa(bnT, bnS, E, ecQ, &digest)
@@ -763,6 +781,10 @@ LIB_EXPORT TPM_RC CryptEccGenerateKey(
 //#  endif										// libtpms changed
     retVal = (OK) ? TPM_RC_SUCCESS : TPM_RC_NO_RESULT;
 Exit:
+    // libtpms added begin: bnD_buf holds the generated ECC private key (the
+    // scalar 'd'); clear it here rather than leaving it in this stack frame.
+    OPENSSL_cleanse(&bnD_buf, sizeof(bnD_buf));
+    // libtpms added end
     CRYPT_CURVE_FREE(E);
     return retVal;
 }
--- a/src/tpm2/TPMCmd/tpm/src/crypt/CryptRsa.c
+++ b/src/tpm2/TPMCmd/tpm/src/crypt/CryptRsa.c
@@ -1549,6 +1549,12 @@ Exit:
         if (retVal)
             retVal = TPM_RC_FAILURE;
     }						// libtpms added end
+    // libtpms added begin: Z holds the generated RSA private key (primes P,
+    // Q and their CRT parameters); clear it here rather than leaving it in
+    // this stack frame. Harmless no-op on the OpenSSL key-generation path,
+    // where Z is never populated.
+    OPENSSL_cleanse(Z, sizeof(*Z));
+    // libtpms added end
     return retVal;
 }
 


src/tpm2/TPMCmd/tpm/src/crypt/CryptEccMain.c
```diff
@@ -754,7 +756,10 @@ LIB_EXPORT TPM_RC CryptEccGenerateKey(
         // Get a random value to sign using the built in DRBG state
         DRBG_Generate(NULL, digest.t.buffer, digest.t.size);
         if(_plat__InFailureMode())
-            return TPM_RC_FAILURE;
+        {
+            retVal = TPM_RC_FAILURE;  // libtpms changed: fall through to Exit
+            goto Exit;                // so bnD is cleared below
```
stefanberger: This should be `ERROR_EXIT(TPM_RC_FAILURE); // libtpms changed`

src/tpm2/TPMCmd/tpm/src/crypt/CryptRsa.c
```diff
@@ -1549,6 +1549,12 @@ LIB_EXPORT TPM_RC CryptRsaGenerateKey(
         if (retVal)
             retVal = TPM_RC_FAILURE;
     }						// libtpms added end
+    // libtpms added begin: _Z holds the generated RSA private key (primes P,
```
stefanberger: `_Z` -> `Z` ?

src/tpm2/TPMCmd/tpm/src/crypt/CryptEccMain.c
Johnsonajibi: Done, thanks -- switched to ERROR_EXIT(TPM_RC_FAILURE) to match the project convention.

src/tpm2/TPMCmd/tpm/src/crypt/CryptRsa.c
Johnsonajibi: Fixed, thanks -- the comment now says Z (matching the OPENSSL_cleanse(Z, sizeof(*Z)) call below it, rather than the internal _Z name from the NEW_PRIVATE_EXPONENT macro).

src/tpm2/TPMCmd/tpm/src/crypt/CryptEccMain.c
```diff
@@ -763,6 +766,10 @@ LIB_EXPORT TPM_RC CryptEccGenerateKey(
 //#  endif										// libtpms changed
     retVal = (OK) ? TPM_RC_SUCCESS : TPM_RC_NO_RESULT;
 Exit:
+    // libtpms added begin: bnD_buf holds the generated ECC private key (the
+    // scalar 'd'); clear it here rather than leaving it in this stack frame.
+    OPENSSL_cleanse(&bnD_buf, sizeof(bnD_buf));
```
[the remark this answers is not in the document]
Johnsonajibi: Good catch -- fixed. Cleared bnExtraBits_buf in both TpmEcc_GenPrivateScalar() variants (it's the raw random value dOut is directly derived from) and bnD1_buf in TpmEcc_GenerateKeyPair() (bnD + public order, which trivially reveals the private scalar bnD -- used for curves whose order isn't at a byte boundary, e.g. P521). Left nMinus1 alone since it's derived only from the curve's public order, not secret material.

Verified live end-to-end this time rather than just a build check: ran swtpm linked against this build through tpm2-tools, TPM2_Startup followed by CreatePrimary for ECC P256 (byte-boundary curve), ECC P521 (non-byte-boundary curve -- specifically exercises the bnD1 path), and RSA (regression check) -- all three succeed.

src/tpm2/TPMCmd/tpm/src/crypt/CryptEccMain.c
```diff
@@ -573,6 +587,11 @@ BOOL TpmEcc_GenerateKeyPair(Crypt_Int*            bnD,  // OUT: private scalar
     } else {
         OK = OK && ExtEcc_PointMultiply(ecQ, NULL, bnD, E);
     }
+    // libtpms added begin: bnD1_buf is bnD + order (order is public), so it
```
stefanberger: You can remove libtpms added being in this case since the whole block is libtpms-specific.

src/tpm2/TPMCmd/tpm/src/crypt/CryptEccMain.c
```diff
@@ -573,6 +587,11 @@ BOOL TpmEcc_GenerateKeyPair(Crypt_Int*            bnD,  // OUT: private scalar
...
+    // trivially reveals the private scalar bnD; clear it here rather than
+    // leaving it in this stack frame.
+    OPENSSL_cleanse(&bnD1_buf, sizeof(bnD1_buf));
+    // libtpms added end
```
stefanberger: Please remove this comment.

src/tpm2/TPMCmd/tpm/src/crypt/CryptRsa.c
```diff
@@ -1549,6 +1549,12 @@ LIB_EXPORT TPM_RC CryptRsaGenerateKey(
         if (retVal)
             retVal = TPM_RC_FAILURE;
     }						// libtpms added end
+    // libtpms added begin: Z holds the generated RSA private key (primes P,
+    // Q and their CRT parameters); clear it here rather than leaving it in
+    // this stack frame. Harmless no-op on the OpenSSL key-generation path,
+    // where Z is never populated.
+    OPENSSL_cleanse(Z, sizeof(*Z));
+    // libtpms added end
```
stefanberger, line 1557: Rest looks good.

src/tpm2/TPMCmd/tpm/src/crypt/CryptEccMain.c
Johnsonajibi: Done, removed.

src/tpm2/TPMCmd/tpm/src/crypt/CryptEccMain.c
Johnsonajibi: Done, removed.

src/tpm2/TPMCmd/tpm/src/crypt/CryptRsa.c
Johnsonajibi: Thanks!

src/tpm2/TPMCmd/tpm/src/crypt/CryptEccMain.c
```diff
@@ -513,6 +521,12 @@ BOOL TpmEcc_GenPrivateScalar(
     OK = OK && ExtMath_Mod(bnExtraBits, nMinus1);
     OK = OK && ExtMath_AddWord(dOut, bnExtraBits, 1);
 
+    // libtpms added begin: bnExtraBits_buf holds the raw random value that
+    // dOut (the private scalar) is directly derived from; clear it here
+    // rather than leaving it in this stack frame.
+    OPENSSL_cleanse(&bnExtraBits_buf, sizeof(bnExtraBits_buf));
+    // libtpms added end
```
stefanberger: Please also remove here.

src/tpm2/TPMCmd/tpm/src/crypt/CryptEccMain.c
```diff
@@ -513,6 +521,12 @@ BOOL TpmEcc_GenPrivateScalar(
     OK = OK && ExtMath_Mod(bnExtraBits, nMinus1);
     OK = OK && ExtMath_AddWord(dOut, bnExtraBits, 1);
 
+    // libtpms added begin: bnExtraBits_buf holds the raw random value that
```
stefanberger: Please also remove here.

src/tpm2/TPMCmd/tpm/src/crypt/CryptEccMain.c
Johnsonajibi: Done, removed both.

src/tpm2/TPMCmd/tpm/src/crypt/CryptEccMain.c
Johnsonajibi: Done, removed both.

src/tpm2/TPMCmd/tpm/src/crypt/CryptEccMain.c
```diff
@@ -478,6 +480,11 @@ BOOL TpmEcc_GenPrivateScalar(
     OK = OK && ExtMath_Mod(bnExtraBits, nMinus1);
     OK = OK && ExtMath_AddWord(dOut, bnExtraBits, 1);
 
+    // bnExtraBits_buf holds the raw random value that dOut (the private
```
stefanberger: Here the two lines need to go back into. Sorry. This is a function provided by the reference code where libtpms now adds a few lines.

Johnsonajibi: No worries, added back -- makes sense, this function is the original reference implementation with just the cleanse added, unlike the other two spots which are fully libtpms-added blocks.

stefanberger:
Thanks!

PR #627, opened 2026-09-15
tpm2: Read attributes from proper NVRAM offset in NvFlushHierarchy

### LINKED ISSUE #626: SRK is persisted when non-null profile in effect
**Describe the bug**

`TPM2_Clear` returns `TPM_RC_SUCCESS`, but owner persistent objects such as the SRK remain present and unchanged. It also survives restarting swtpm with the same state.

**To Reproduce**

Steps to reproduce the behavior/issue showing all commands on command line, needed XML or JSON (if necessary), etc.:

1. Manufacture fresh state with `swtpm_setup --tpm2 --create-spk`.
2. Start swtpm with `--flags not-need-init,startup-clear`.
3. Verify persistent handle `0x81000001` exists.
4. Send `TPM2_Clear` authorized by `TPM_RH_LOCKOUT`.
5. Query persistent handles and `TPM2_ReadPublic(0x81000001)`.
6. Restart swtpm and query again.

**Expected behavior**

Persistent handles before Clear: [0x81000001]
TPM2_Clear response: 0x00000000
Persistent handles after Clear:  []

**Observed**

- `TPM2_Clear` response: `TPM_RC_SUCCESS`
- `0x81000001` remains present immediately after Clear.
- Its `TPM2_ReadPublic` response is byte-identical.
- It remains present after restart.

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Version 26.04

**Versions of relevant components**
- libtpms: 0.10.2

**Additional context**

For `StateFormatLevel >= 2`, `NvObjectToBuffer()` stores persistent objects using `ANY_OBJECT_Marshal()`.

However, `NvFlushHierarchy()` still reads `OBJECT_ATTRIBUTES` using:

https://github.com/stefanberger/libtpms/blob/18e42c8deec110ba8fdf1168f63204e993d90a83/src/tpm2/TPMCmd/tpm/src/subsystem/NvDynamic.c#L1431-L1433

That offset is only valid for the old raw OBJECT layout. Consequently, spsHierarchy/epsHierarchy are not detected and the persistent object is not deleted.

The recommended fix is to read the object through `currentAddr` and derive the attributes from the object.

stefanberger:
The attributes of an OBJECT were previously at a fixed offset in the NVRAM byte array. However, since ANY_OBJECT_Marshal is now being used to write an OBJECT into the NVRAM, the offset may be different depending on the current StateFormatLevel and the layout of the OBJECT. Therefore, ANY_OBJECT_Unmarshal has to be used to get the attributes of an OBJECT.

Fixes: #626

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of evicted objects when flushing a hierarchy, ensuring their attributes are loaded through the standard object-reading process.


kuqin12 (approved):

stefanberger:
Note: Needs to be backported to v0.10 but not earlier.

PR #633, opened 2026-09-17
tpm2: Add an assert to avoid integer underrun

stefanberger:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved stability when reading malformed or undersized stored data.
  - Prevented invalid size calculations that could cause failures during NVRAM object reads.
  - Corrected NVRAM error reporting so write-failure offsets are displayed accurately across supported platforms.


danieltrick:
Wroks for me 👍

